PR #20, opened 2022-04-05
Added support for chargers that are not reachable as installations

mrfrostfire:
This adds support for PRO chargers that does not turn up as installations, and also removes some deprecation-warnings

### DIFF
--- a/custom_components/zaptec/switch.py
+++ b/custom_components/zaptec/switch.py
@@ -17,13 +17,13 @@ async def async_setup_platform(
     hass: HomeAssistantType, config: ConfigType, async_add_entities, discovery_info=None
 ) -> bool:  # pylint: disable=unused-argument
     """Setup switch platform."""
-    api = hass.data.get(DOMAIN, {}).get("api")
-    if api is None:
+    acc = hass.data[DOMAIN]["api"]
+    if acc is None:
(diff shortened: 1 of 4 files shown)


custom_components/zaptec/switch.py
```diff
@@ -17,13 +17,13 @@ async def async_setup_platform(
     hass: HomeAssistantType, config: ConfigType, async_add_entities, discovery_info=None
 ) -> bool:  # pylint: disable=unused-argument
     """Setup switch platform."""
-    api = hass.data.get(DOMAIN, {}).get("api")
-    if api is None:
+    acc = hass.data[DOMAIN]["api"]
```
Hellowlol: Use .get like I did before. This will raise and exception and not return None

mrfrostfire:
Ok i’ll change that back. I changed it to this because In anorher file you did it this way

custom_components/zaptec/switch.py
mrfrostfire: fixed

mrfrostfire:
Thanks :)

Olen:
Any chance of getting this in a new release?
Anyone verified it with HOME chargers?

I have tested the patch on our PRO-chargers and it works fine for me.

Hellowlol:
I haven’t had time to test it yet. I’ll try to get it tested and released this weekend

PR #39, opened 2023-07-21
Inline

Hellowlol:


custom_components/zaptec/sensor.py
```diff
@@ -78,20 +90,39 @@ async def async_setup_entry(hass, config_entry, async_add_devices):
...
+    async def async_added_to_hass(self):
+        """Connect to dispatcher listening for entity data notifications."""
+        await super().async_added_to_hass()
+        _LOGGER.debug("called async_added_to_hass %s", self.__class__.__name__)
+        self.async_on_remove(
+            async_dispatcher_connect(self._hass, EVENT_NEW_DATA, self._real_update)
+        )
+
+    @property
+    def should_pull(self):
```
sveinse, line 113: @Hellowlol `should_pull(self)` is a typo. It should be `should_poll(self)`.

PR #41, opened 2023-07-22
Fix the spelling to "circuit"

sveinse:
This PR is a minor fix that renames the spelling of "Circute" to "Circuit". It will affect the `zaptec_circuit_*` entity name.


Hellowlol:
Thanks!

Closing this as a fix is already included in https://github.com/custom-components/zaptec/pull/39

PR #42, opened 2023-07-22
Add support for reading device type schemas

sveinse:
The Zaptec constants JSON contains in addition to the global generic id mappings, contain device specific schemas. This PR adds support for appending the device type specific ids list with the generic id list. It will significantly reduce the number of `Couldnt find a remap string for %s report it %s"` messages in the log.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -53,7 +53,7 @@ class AuthorizationFailedException(Exception):
 
 
 # should be a static method of account
-async def _update_remaps() -> None:
+async def _update_remaps(device_types=None) -> None:
     wanted = ["Observations"]
     obs = {}
     async with aiohttp.request("GET", CONST_URL) as resp:
@@ -64,6 +64,15 @@ async def _update_remaps() -> None:
                     obs.update(v)
                     # Add names.
                     obs.update({value: key for key, value in v.items()})
+
+                # Get from Schema.<Name>.ObservationIds for those where
+                # Schema.<Name>.DeviceType matches the list in device_types
+                if device_types and k == "Schema":
+                    for schema in v.values():
+                        v2 = schema.get("ObservationIds")
+                        if schema.get("DeviceType") in device_types and v2 is not None:
+                            obs.update(v2)
+                            obs.update({value: key for key, value in v2.items()})
 
         _LOGGER.debug("Update remaps")
         return obs
@@ -440,8 +449,6 @@ class Account:
 
     async def build(self):
         """Make the python interface."""
-        if not len(self.obs):
-            self.obs = await _update_remaps()
         installations = await self.installations()
 
         cls_installs = []
@@ -456,10 +463,19 @@ class Account:
         self.installs = cls_installs
 
         so_chargers = await self.chargers()
+        device_types = []
         for charger in so_chargers:
             if charger.id not in self.map:
                 self.map[charger.id] = charger
+
+            # Append the device type in order to read the extended
+            # IDs from the remap dict.
+            device_types.append(charger.device_type)
+
         self.stand_alone_chargers = so_chargers
+
+        if not len(self.obs):
+            self.obs = await _update_remaps(device_types)
 
 
 class Charger(ZapBase):


Hellowlol:
Hmm, how does this if a user has multiple chargers that they have access to? Lets say one new one and one old?

custom_components/zaptec/api.py
```diff
@@ -447,11 +454,21 @@ async def build(self):
...
 
         so_chargers = await self.chargers()
+        device_types = []
         for charger in so_chargers:
             if charger.id not in self.map:
                 self.map[charger.id] = charger
+
+            # Append the device type in order to read the extended
+            # IDs from the remap dict.
+            device_types.append(charger.device_type)
```
sveinse, line 473: It collects all present chargers in order to determine which device specific schemas to include.

Hellowlol:
We are still just pushing data into self.obs

Not all the observation has the same key/value:
key ChargeCurrentInstallationMaxLimit Observations 707 Apollo 546

Can you check this one? I don't think its that important. As long as we add the value as a key so we know the correct attribute to update if the user is using the servicebus

sveinse:
The data in self.obs are only used one way, right? Even though the dict is populated for two way use (`obs[k] = v; obs[v] = k`). My assumptions have been that two different devices might use two different IDs for the same thing. I hope that the same ID won't be reused for a completely different thing in another device (collision). If that's the case we'll need to have have device specific ID-lists.

I'll check for collisions with the charger I have.

sveinse:
I made the following snippet shown below to find if there are names using multiple IDs and IDs having multiple names. The findings for _all_ device schemas included in the const file gives the following overlap. They seem pretty related to me, which I think is good for doing it this way.

```
Found ChargeCurrentInstallationMaxLimit with multiple values {546, 707}
Found 202 with multiple values {'TemperatureEMeterPhase1', 'TemperatureInternal6'}
Found 205 with multiple values {'TemperaturePowerBoard', 'TemperatureEMeterPhase3'}
Found 522 with multiple values {'ChargerOfflinePhase', 'VoltageEnergyLevel'}
Found 813 with multiple values {'McuToVariscitePacketErrors', 'McuToESPPacketErrors'}
Found 814 with multiple values {'ESPToMcuPacketErrors', 'VarisciteToMcuPacketErrors'}
Found 820 with multiple values {'UptimeESP', 'UptimeVariscite'}
Found 911 with multiple values {'SoftwareApplicationVersion', 'SmartComputerSoftwareApplicationVersion'}
Found 913 with multiple values {'HardwareVersion', 'SmartComputerHardwareVersion'}
Found 970 with multiple values {'FactoryTestStage', 'ProductionTestStationNumber'}
Found 1000 with multiple values {'MeterId', 'AMSMeterId'}
Found 1001 with multiple values {'AMSMeterType', 'MeterType'}
```

```py
    stats = {}
    async with aiohttp.request("GET", CONST_URL) as resp:
        if resp.status == 200:
            data = await resp.json()
            for k, v in data.items():
                if k in wanted:
                    for a, b in v.items():
                        stats.setdefault(a, []).append(b)
                        stats.setdefault(b, []).append(a)

                # Get from Schema.<Name>.ObservationIds for those where
                # Schema.<Name>.DeviceType matches the list in device_types
                if device_types and k == "Schema":
                    for schema in v.values():
                        if 'ObservationIds' in schema:
                            for a, b in schema['ObservationIds'].items():
                                stats.setdefault(a, []).append(b)
                                stats.setdefault(b, []).append(a)

        for k, v in stats.items():
            sv = set(v)
            if len(sv) > 1:
                _LOGGER.error("Found %s with multiple values %s", k, sv)
```

Hellowlol:
> The data in self.obs are only used one way, right? Even though the dict i

No, its is not, it allows two-way because the data from the servicebus only contains the StateID

sveinse:
> > The data in self.obs are only used one way, right? Even though the dict i
> 
> No, its is not, it allows two-way because the data from the servicebus only contains the StateID

Yes, everything from the API use numerical IDs and needs translation into string. But the opposite direction from name to ID, is that ever used?

sveinse:
This PR can be closed if PR #56 is implemented.

sveinse:
Superseded by #56. Closing PR

Issue #44, opened 2023-07-23
Unloading or disabling zaptec will crash it


sveinse:
Unloading zaptec will create tracebacks, which in turn will result in the server having to be restarted. This happens on uninstall and on disabling.

```
2023-07-22 22:41:17.335 ERROR (MainThread) [homeassistant.config_entries] Error unloading entry Zaptec for sensor
Traceback (most recent call last):
  File "/workspaces/ha-core/homeassistant/config_entries.py", line 547, in async_unload
    result = await component.async_unload_entry(hass, self)
  File "/workspaces/ha-core/homeassistant/components/sensor/__init__.py", line 140, in async_unload_entry
    return await component.async_unload_entry(entry)
  File "/workspaces/ha-core/homeassistant/helpers/entity_component.py", line 202, in async_unload_entry
    raise ValueError("Config entry was never loaded!")
ValueError: Config entry was never loaded!
```

PR #45, opened 2023-07-23
Fix issues with crash when unloading or disabling zaptec

sveinse:
Unloading the zaptec integration will create a trace-back in the logs, see #44. This PR proposes a fix for it. 

The `_stream()` task was blocking the cleanup which is the root cause for the crash. The fix sends a `close()` to the azure-stream object and then cancels the stream task. That seems to work fine in my installation.

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -71,24 +71,13 @@ async def async_setup(hass: HomeAssistantType, config: ConfigType) -> bool:
 async def async_setup_entry(hass: HomeAssistantType, entry: ConfigEntry) -> bool:
     """Set up zaptec as config entry."""
     await _dry_setup(hass, entry.data)
-
-    for platform in PLATFORMS:
-        hass.async_create_task(
-            hass.config_entries.async_forward_entry_setup(entry, platform)
-        )
+    await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
     return True
 
 
 async def async_unload_entry(hass: HomeAssistantType, entry: ConfigEntry):
     """Unload a config entry."""
-    unload_ok = all(
-        await asyncio.gather(
-            *[
-                hass.config_entries.async_forward_entry_unload(entry, component)
-                for component in PLATFORMS
-            ]
-        )
-    )
+    unload_ok = await hass.config_entries.async_unload_platforms(entry, PLATFORMS)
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/__init__.py
```diff
@@ -71,24 +71,13 @@ async def async_setup(hass: HomeAssistantType, config: ConfigType) -> bool:
 async def async_setup_entry(hass: HomeAssistantType, entry: ConfigEntry) -> bool:
     """Set up zaptec as config entry."""
     await _dry_setup(hass, entry.data)
-
-    for platform in PLATFORMS:
-        hass.async_create_task(
```
Hellowlol, line 82: I'm not sure this is a proper fix, im memory serves me there was some issue with import if we didn't used async_create_task.

sveinse: The only difference between the old implementation and the new, is that the latter runs `asyncio.gather()` on the execution of the tasks. The old code simply creates the tasks and leaves them to their own devices. A hang here will hang the setup in the new implementation.

(The new is AFAICS the preferred way to setup the platforms)

custom_components/zaptec/api.py
```diff
@@ -295,6 +297,7 @@ async def _stream(self, cb=None):
     async def cancel_stream(self):
         if self._stream_task is not None:
             try:
+                await self._receiver.close()
```
Hellowlol: Can try do it more defensively? Check that self._receiver is not and check that is has a close method

custom_components/zaptec/__init__.py
Hellowlol: Sweet!

custom_components/zaptec/api.py
sveinse: Yes, that is probably a good idea. When the stream task is waiting on `async for msg in receiver:`, it will hang when the stream task is canceled and awaited on. It must relinguish this loop and calling `self._receiver.close()` worked, albeit a bit brute force.

sveinse: Updated to more defensive coding. Added try...finally to ensure the receiver object is only valid during reception and check that its not empty before calling `close()`.

I didn't add a check for close thou. Isn't it ok to assume that the servicebus client receiver indeed has a `close()` method as its docs claim?

PR #51, opened 2023-07-25
fix reg

Hellowlol:


sveinse:
Looks good (haven't checked it on HW yet). 

While fixing these things, might I suggest integrating a couple of some additional minor fixes, see https://github.com/sveinse/zaptec/commit/3897b71c6e423ee56ba41d7145248ddc13222246

I can add them as a PR if you'd rather want that.

sveinse:
Can we close this issue? 0.7.0 beta has been released and I'm not sure this PR is needed any more.

PR #53, opened 2023-07-25
Warnings in logs of exceeding db max size

sveinse:
I keep getting the following warning in HA logs:

```
[homeassistant.components.recorder.db_schema] State attributes for sensor.zaptec_installation_xxx exceed maximum size of 16384 bytes. This can cause database performance issues; Attributes will not be stored
```

This is caused by an encoded icon as a parameter (`SupportGroup.LogoBase64`) in the GET of `installation/xxx`. It is 28800 bytes. This PR replaces this excessive attribute with the text "<Removed, was 28800 bytes>"

The PR also adds the attrs `ChargerMinCurrent` and `ChargerMaxCurrent` to the charger sensor.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -438,6 +438,16 @@ class Account:
 
     async def installation(self, installation_id):
         data = await self._request(f"installation/{installation_id}")
+
+        # Remove data fields with excessive data, making it bigger than the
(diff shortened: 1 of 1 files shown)


Hellowlol:
Thanks, i havnt noticed as the charger i have access to don't have such a value

PR #93, opened 2024-01-22
More about load balancing, better formatting.

svenakela:
The contribution about load balancing was messing up the readme, I moved some text and added more info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,28 +170,6 @@ The service call `limit_current` can be used with the arguments
 `available_current_phase3` to set the available current on individual phases.
 
 
-## Load balancing your charger
-
-By using the [Zaptec Load Balancing](https://github.com/svenakela/ha/tree/main/zaptec) 
-blueprint you'll get automatic load balancing for your charger (i.e. the charger 
-limit is updated constantly to avoid fuse overload).
-
-The automation created with the blueprint manages current limiting only. It will check 
-the charger status and sets the current accordingly. If charging is enabled and 
-possible it will manage the limit. It will _not_ maintain the charging nor scheduling.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -162,26 +162,35 @@ changing this value will affect all.
...
-blueprint you'll get automatic load balancing for for your charger (i.e. the 
-charger limit is updated constantly to avoid fuse overload).
+blueprint you'll get automatic load balancing for your charger (i.e. the charger 
+limit is updated constantly to avoid fuse overload).
+
+The automation created with the blueprint manages current limiting only. It will check 
+the charger status and sets the current accordingly. If charging is enabled and 
+possible it will manage the limit. It will _not_ maintain the charging nor scheduling.
+
+How setup the automation, how logic works and what all settings mean is documented in the [blueprint readme](https://github.com/svenakela/ha/blob/main/zaptec/README.md).
```
philsson: How ***to*** setup...

sveinse:
I have one concern: I think its important that the features or implementations of the load balancing blueprint is not tied too tightly with this project. If we do, this will result in this repo having to update whenever new features were changed in the blueprint repo. So I think the readme should give very general information that the load balancing blueprint exists and then refer to the blueprint repos readme.

svenakela:
> I have one concern: I think its important that the features or implementations of the load balancing blueprint is not tied too tightly with this project. If we do, this will result in this repo having to update whenever new features were changed in the blueprint repo. So I think the readme should give very general information that the load balancing blueprint exists and then refer to the blueprint repos readme.

I don't mind, feel free to remove as much as you like. Maybe even move to the bottom of the readme as "Usage Examples" or similar?

svenakela:
@sveinse had some spare time to adjust the readme. I've shortened the text and moved it to the bottom of the file.

svenakela:
@sveinse is this PR stale? The updates are at least much better than the previously approved adds.

PR #111, opened 2024-08-08
Retry on API error code 500 (#90)

### LINKED ISSUE #90: Appropriate retry mechanism when Zaptec cloud is not working
Discussion: What is the appropriate action when Zaptec cloud is not responding or returns error?

Currently the Zaptec integration will log them to the HA system log. This can easily be interpreted that its the zaptec integration which is at fault. Typical errors that are observed are timeouts and HTTP error code 500 (internal server error). When this happens, quite many errors and warnings are logged.

The [Home Assistant integration Quality scale](https://developers.home-assistant.io/docs/integration_quality_scale_index/) states that a integration should (for silver):
* Handles internet unavailable. Log a warning once when unavailable, log once when reconnected.
* Handles device/service unavailable. Log a warning once when unavailable, log once when reconnected.

So what should we do? Do we need to implement some way to determine Zaptec cloud's "healthyness" in order for our integration to reduce the error logging?

sveinse:
Add retries if error code 500 is received. Fixes #90

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.7.2b1
 
+* Add retry on 500 server errors from Zaptec cloud, #90
 * Remove large and noisy charger states containg test or production data.
   Helps to reduce the entity attribute <16kb. #88
 * Ignore updates with chargerid 000000...000, #84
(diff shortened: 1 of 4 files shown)


sveinse:
As described in #90 , a slightly more advanced retry mechanism are needed. An update to the PR is required

PR #134, opened 2024-10-25
Dutch language

c0mplex1:
This file contains the Dutch language

### DIFF
--- a/nl.json
+++ b/nl.json
@@ -1,0 +1,132 @@
+{
+    "config": {
+        "step": {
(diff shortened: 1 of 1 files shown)


sveinse (changes requested):
You're doing great on making a PR. I've read through the file and it looks good. (Despite being Norwegian, I am able to follow much of the words.)

Can you please move the file to custom_components/zaptec/translations/nl.json, please? This is the location which the file needs to reside for it to work.

PR #140, opened 2024-10-28
Polish language

emce:
Added Polish translation based on newest English one.


emce:
Sure - please subscribe me for notifications.

PR #141, opened 2024-10-29
Swedish translation

FluGanz:
Added a swedish translation


sveinse:
Approved.

@FluGanz is it ok to add you on a list to update the Swedish translations for new changes that needs translations?

PR #145, opened 2024-11-20
Total -> Total Increasing

tleegaard:
As I understand the documentation at https://developers.home-assistant.io/blog/2021/08/16/state_class_total/ and https://developers.home-assistant.io/docs/core/entity/sensor/#available-state-classes these sensors should be TOTAL_INCREASING as they're decreasing at the end of the session.

### DIFF
--- a/custom_components/zaptec/sensor.py
+++ b/custom_components/zaptec/sensor.py
@@ -216,7 +216,7 @@ CHARGER_ENTITIES: list[EntityDescription] = [
         device_class=SensorDeviceClass.ENERGY,
         icon="mdi:counter",
         native_unit_of_measurement=const.UnitOfEnergy.KILO_WATT_HOUR,
-        state_class=SensorStateClass.TOTAL,
+        state_class=SensorStateClass.TOTAL_INCREASING,
     ),
     ZapSensorEntityDescription(
         key="signed_meter_value_kwh",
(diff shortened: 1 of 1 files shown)


sveinse:
There were an update to the blog here: https://developers.home-assistant.io/blog/2021/09/20/state_class_total/ that states that `TOTAL` is intended for values that can both increase and decrease, while `TOTAL_INCREASING` is for sensors that monotonically increase.  I know there were some confusion about this a while back, but I _think_ `TOTAL` is the right choice here, since the two sensors in this PR jumps up and down based on the current session.

Is there a specific reason why you make this proposal? Is there something that affects your usage of Zaptec by having `TOTAL` here? I knew we had some issues with integrating with the Energy view in HA, but according to the docs https://next.home-assistant.io/docs/energy/faq/#troubleshooting-missing-entities having either `TOTAL` or `TOTAL_INCREASING` should work.

tleegaard:
My specific reason is that due to the state_class being 'total', the sum is getting negative values, as described at the blog you're linking: https://developers.home-assistant.io/blog/2021/09/20/state_class_total/
And when using the sensor as individual device in the energy dashboard, the values increase while charging and decrease when the car is unplugged:
image
As I understand the 'total_increasing' state_class, that would fix the problem 😊

sveinse:
Yes, looking closer at the same document I linked to and inspecting the HA code, I agree and support the fix. ...HA haven't exactly been very clear about this in the past. Let's hope we can finally settle it

sveinse (approved):

Issue #176, opened 2025-05-14
Calling internal GET /api/chargers/[id]/settings


ivastokic:
Hello, Zaptec devs are here!

We’ve noticed that your integration is making sequential calls to the following endpoints:

`GET api/chargers/[id]/state`

`GET api/chargers/[id]/settings`

Just a quick note: the `state` endpoint already includes all the data returned by the `settings` endpoint — in fact, `settings` is simply a subset of `state`. While you might occasionally see slight differences in overlapping fields, this typically happens when a charger hasn’t updated its values yet. Once it does, the accurate value will be reflected in state.

Additionally, please note that the settings endpoint is considered internal. It is subject to change or rate limiting without prior notice, and we recommend avoiding reliance on it for production use.

We kindly ask that you remove the settings call (https://github.com/custom-components/zaptec/blob/master/custom_components/zaptec/api.py#L604-L613) if you're already using state, because we have plans to forbid usage of this endpoint by external parties.

We truly appreciate your support in optimizing the integration. If you have any questions or need help making this adjustment, feel free to reach out.

Thanks again for your cooperation!

sveinse:
> The other place we noticed that needs improvement is sending of `Pause/Resume` charging, no matter of state charger is in (for example, `Disconnected` state). In general, `Pause` should be sent only if charger is `Charging`, and `Resume` only if charging is `Paused`. But we will come back to you with more details once we revisit other places.

It would be nice if you could add the issues for findings like this. Home Assistant is rather stateless, so this will require some system thinking into the overall architecture. It would be nice to get all those inputs before going into a bigger rewrite.

sveinse:
> Just a quick note: the `state` endpoint already includes all the data returned by the `settings` endpoint — in fact, `settings` is simply a subset of `state`. While you might occasionally see slight differences in overlapping fields, this typically happens when a charger hasn’t updated its values yet. Once it does, the accurate value will be reflected in state.

I tested this on my system at home and I found three fields that are not present anywhere other than in settings.

```json
["CurrentInMaximum", "DefaultOfflineCurrent", "CurrentInMinimum", "DefaultOfflinePhase"]
```

This shows _are_ information in settings which is not reflected in the charger info and state endpoint. The use case for `CurrentInMinimum` and `CurrentInMaxiumum` is to be able to configure the current range on the charger. In order to have a control for them, the current value must be read, and the settings endpoint provides that.

[...]

PR #183, opened 2025-06-29
Remove internal settings api calls

steinmn:
Addresses https://github.com/custom-components/zaptec/issues/176

Unless I've overlooked something, the GET call to settings is not actually used for anything, since the values are already covered by the GET call to state.

The POST call to settings is replaced with the official update endpoint https://docs.zaptec.com/reference/charger_updateexternal_post. (Currently getting a validation warning, so will look into handling that before marking the PR as ready)

Had to update the requirements to get my dev-container running, let me know if you would prefer this as a separate PR @sveinse


sveinse:
I think "threshold for new devs" is how its said in English. I think. Not all Norwegian expressions are translatable and the other way round.

Anyhow, I agree, the format on save should be disabled. Do you mind updating #184 while at it? In my mind it falls to somewhat the same category. Also agree about doing a writeup of the developer experience. Let's open an issue about that and continue that conversation in that, please.

I'll take a look at the PR.

sveinse:
At an initial read-through this looks good. I'd like to test this for a few days on my test setup. Unfortunately my car is just recently topped up, so I can't test a charge cycle just now. But I'll keep it running for the days to come and run the next charge cycle with it.

sveinse:
The fix seems to be working fine. Great work.

I have a UX issue that's due to how this integration is polling its data. If I set the "Charger max current" from 32A to 12A, it will jump back to 32A after a few seconds. After a minute it will jump to the final value of 12A. I did some debugging and found it to be cause by the following sequence. Logs below.

1. Value of `maxChargeCurrent` is set from 32A to 12A
2. Service bus immediately responds and updates the entity to 12A
3. HA polls Zaptec cloud and gets the old value of 32A
4. On next poll interval when HA polls Zaptec, it gets update back to 12A

Here we are touching a fundamental problem: Whenever a entity is updated by HA, it polls the Zaptec cloud for updates. This is necessary if changing a value or pressing a button changes other values than the button or number. However, it turns out that the Zaptec cloud is slow at updating its value, which leads to this ping-pong of value.

If one disables `await self.coordinator.async_request_refresh()` in number.py#L95 this issue will be reduced, as it does not initiate a cloud pull immediately after changing the value. However, if the integration is scheduled to update its value very soon after the number update, the user will see a value change back to the original value which will last the length of the poll interval.


```sh
# Setting to 12A
26:20.450 [zaptec.number] Setting ZaptecSettingNumber.charger_max_current to <float> 12.0   (in ...f40861)
26:20.451 [zaptec.api] Settings {'maxChargeCurrent': 12.0}

# Service bus immediately responds with the new value
26:21.562 [zaptec.api] ---   Subscription: {'DeviceId': 'xxx', 'DeviceType': 4, 'ChargerId': '...f40861', 'StateId': '510 (ChargerMaxCurrent)', 'Timestamp': '2025-07-06T09:26:21.065669Z', 'ValueAsString': '12.000'}
26:21.562 [zaptec.api] >>>   Updating Charger[f40861].charger_max_current (ChargerMaxCurrent)  =  <float> 12.0  (was 32.0)
26:21.563 [zaptec]     number.sol_charger_max_current  =  <float> 12.0   from Charger[f40861]
26:21.597 [zaptec.api] ---   Subscription: {'DeviceId': 'xxx', 'DeviceType': 4, 'ChargerId': '...f40861', 'StateId': '522 (ChargerOfflinePhase)', 'Timestamp': '2023-07-12T14:10:37.23', 'ValueAsString': '4'}
26:21.621 [zaptec.api] ---   Subscription: {'DeviceId': 'xxx', 'DeviceType': 4, 'ChargerId': '...f40861', 'StateId': '712 (IsStandAlone)', 'Timestamp': '2023-07-12T13:31:40.253', 'ValueAsString': '0'}

# HA polls for new updated values as a response to the setting
# -- but cloud is not updated yet so the value jumps back to 32A
26:21.664 [zaptec.api] Polling state for Installation[8a106b] (Sol Installation)
26:21.768 [zaptec.api] Polling state for Circuit[84a4e5] (Sol Circuit)
26:21.822 [zaptec.api] Polling state for Charger[f40861] (Sol)
26:21.924 [zaptec.api] >>>   Updating Charger[f40861].charger_max_current (ChargerMaxCurrent)  =  <float> 32.0  (was 12.0)
26:21.987 [zaptec] Finished fetching xxx data in 0.322 seconds (success: True)
26:21.988 [zaptec]     number.sol_charger_max_current  =  <float> 32.0   from Charger[f40861]

# In the next scheduled the poll, the value is updated to 12A
27:21.458 [zaptec.api] Polling state for Installation[8a106b] (Sol Installation)
27:21.628 [zaptec.api] Polling state for Circuit[84a4e5] (Sol Circuit)
27:21.678 [zaptec.api] Polling state for Charger[f40861] (Sol)
27:21.769 [zaptec.api] >>>   Updating Charger[f40861].charger_max_current (ChargerMaxCurrent)  =  <float> 12.0  (was 32.0)
27:21.829 [zaptec] Finished fetching xxx data in 0.372 seconds (success: True)
27:21.831 [zaptec]     number.sol_charger_max_current  =  <float> 12.0   from Charger[f40861]
```

steinmn:
Yeah, now that you mention it, I've seen similar flipflopping with the Charging switch as well.

Considering this is a more fundamental issue, should we try to do anything about it in this PR, or should it be merged as-is and open a discussion on a separate "Polling outdated data"-issue?

sveinse (approved):
Yes, I was thinking the same that I think this PR is good and we can move on. I'm typing up a comment to the thread with Zaptec devs to discuss this principally.

PR #186, opened 2025-06-30
Remove internal circuits api calls

steinmn:
The installation/{self.installation.id}/hierarchy api contains a list of circuits associated with an installation, and the list elements contain the same information as the circuits/{id} api.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -214,7 +214,7 @@ class ZaptecBase(ABC):
 class Installation(ZaptecBase):
     """Represents an installation"""
 
-    circuits: list[Circuit]
+    chargers: list[Charger]
 
     # Type conversions for the named attributes (keep sorted)
     ATTR_TYPES = {
@@ -228,7 +228,7 @@ class Installation(ZaptecBase):
     def __init__(self, data, account):
         super().__init__(data, account)
         self.connection_details = None
-        self.circuits = []
+        self.chargers = []
 
         self._stream_task = None
         self._stream_receiver = None
@@ -247,19 +247,23 @@ class Installation(ZaptecBase):
                     "Access denied to installation hierarchy of %s. The user might not have access.",
                     self.id,
                 )
-                self.circuits = []
+                self.chargers = []
                 return
             raise
 
-        circuits = []
-        for item in hierarchy["Circuits"]:
-            _LOGGER.debug("    Circuit %s", item["Id"])
-            circ = Circuit(item, self._account, installation=self)
-            self._account.register(item["Id"], circ)
-            await circ.build()
-            circuits.append(circ)
+        chargers = []
+        for circuit_item in hierarchy["Circuits"]:
+            _LOGGER.debug("    Circuit %s", circuit_item["Id"])
+            for charger_item in circuit_item["Chargers"]:
+                _LOGGER.debug("      Charger %s", charger_item["Id"])
+                charger_item["CircuitId"] = circuit_item["Id"]
+                charger_item["CircuitName"] = circuit_item["Name"]
+                chg = Charger(charger_item, self._account, installation=self)
+                self._account.register(charger_item["Id"], chg)
+                await chg.build()
+                chargers.append(chg)
 
-        self.circuits = circuits
+        self.chargers = chargers
 
     async def state(self):
         _LOGGER.debug(
@@ -473,57 +477,6 @@ class Installation(ZaptecBase):
         return result
 
 
-class Circuit(ZaptecBase):
-    """Represents a circuits"""
-
-    chargers: list["Charger"]
-
-    # Type conversions for the named attributes (keep sorted)
-    ATTR_TYPES = {
-        "active": bool,
-    }
-
-    def __init__(
-        self, data: TDict, account: Account, installation: Installation | None = None
-    ):
-        super().__init__(data, account)
-        self.chargers = []
-        self.installation = installation
-
-    async def build(self):
-        """Build the python interface."""
-
-        chargers = []
-        for item in self._attrs["chargers"]:
-            _LOGGER.debug("      Charger %s", item["Id"])
-            chg = Charger(item, self._account, circuit=self)
-            self._account.register(item["Id"], chg)
-            await chg.build()
-            chargers.append(chg)
-
-        self.chargers = chargers
-
-    async def state(self):
-        _LOGGER.debug(
-            "Polling state for %s (%s)", self.qual_id, self._attrs.get("name")
-        )
-        data = await self.circuit_info()
-        self.set_attributes(data)
-
-    # -----------------
-    # API methods
-    # -----------------
-
-    async def circuit_info(self) -> TDict:
-        """Obtain circuit data through the installation-hierarchy api"""
-        if self.installation:
-            data = await self._account._request(f"installation/{self.installation.id}/hierarchy")
-            for circuit in data["Circuits"]:
-                if circuit["Id"] == self.id:
-                    return circuit
-        return {}
-
-
 class Charger(ZaptecBase):
     """Represents a charger"""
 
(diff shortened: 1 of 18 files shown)


sveinse:
Let's take one step back: Should we rather remove the `Circuit` object altogether? The only entity provided by the circuit device is `Max current`, and I'm not sure if anyone actually uses this value for anything. If Zaptec is moving away from the circuit API, then I think we should follow suit?

The only time `Circuit.MaxCurrent` is different than `Installation.MaxCurrent` is for those users that have multiple circuits. As you point out, I'm uncertain how common they are (for the use case when used with HA).

steinmn:
Hmmm... That does sound tempting.

We haven't received any indication that they will be removing the Circuit-level from their system hierarchy, have we? Just that we shouldn't use the direct circuit API? Otherwise that would break the Installation.build function (and probably some other stuff) as well.

I'll do a little bit of testing of what it would look like to remove it.

sveinse:
If circuit doesn't provide us with any useful value, then I don't think we need to include it. It's not like we _have_ to model this equal to Zaptecs data model.

`installation/{id}/hiearchy` will still show the circuits and is used to detect the chargers connected to them. I think I would collect all chargers in a list in the installation object instead. Then we can remove the `Charger` class and the updates of it. Possibly store the circuit ID in the charger class. `Charger.circuit` would need to be changed to `Charger.installation` or some such.

Looking forward to your testing

steinmn:
I ended up adding a diagnostic sensor with the circuit name for each charger, so in a way the hierarchy is still included.

The main issue right now is that the old Circuit device is still included in the list of devices connected to the integration. Don't know if this is something we can fix from the backend, or if we need to include some user instructions to fix this.

steinmn:
@feliciaan: As a user with multiple chargers, it would be nice to get some feedback from you as well on this.

Does the Circuit-device(s) and its associated entities provide any kind of value to you in your setup? Or do you agree that it makes sense to simplify the HA architecture down to just Installation- and Charger-devices?

sofa74surfer:
We have an Installation with 4 Circuits. 63A per Circuit and 125A per Installation. Actually we have 10 Chargers (1*3, 1*3, 1*1, 1*3). For Automation, the Information about Circuits isn't important for us. We do not use anything from Circuit.  I can confirm the conclusion in https://github.com/custom-components/zaptec/pull/186#issuecomment-3062899335. We do not have any value, about Ciruits.

sveinse:
@steinmn I usually prefer first-come-first-serve for merges, and this one is first in line. I will be working on this integration today, so we can aim on getting it merged tonight. I'll take a deeper look at it. I hope its ok if I commit a few improvements to your PR.

feliciaan:
@steinmn I can see a use case for the circuits, when you have certain limits per circuit because it has some other loads as well, so that the circuits can be throttled, but I think that this is out-of-scope for this integration. We don't use them for the moment as most of our circuits are directly connected to the grid.

custom_components/zaptec/api.py
```diff
@@ -531,6 +487,8 @@ class Charger(ZaptecBase):
         "charger_max_current": float,
         "charger_min_current": float,
         "charger_operation_mode": ZCONST.type_charger_operation_mode,
+        "circuit_id": str,
+        "circuit_name": str,
```
sveinse: Strictly speaking, attributes that are string doesn't need to be listed here. But it doesn't hurt.

custom_components/zaptec/api.py
```diff
@@ -1107,6 +1065,9 @@ def get_chargers(self):
         """Return a list of all chargers"""
         return [v for v in self.map.values() if isinstance(v, Charger)]
 
+    def get_circuit_id_set(self):
+        return set(c.circuit_id for c in self.get_chargers() if "circuit_id" in c._attrs and len(c.circuit_id)>0)
```
sveinse: First I don't like the name. Lose the `_set` postfix. Please add docstring and type annotation to the function. Next, the function is more complex than it needs to be I think.  So something like this:

```python
    def get_circuit_ids(self) -> set[str]:
        """Return a set of all circuit ids for the chargers."""
        return set(c.circuit_id for c in self.get_chargers() if c.get("circuit_id", ""))
```

custom_components/zaptec/config_flow.py
```diff
@@ -138,11 +138,10 @@ async def async_step_chargers(
...
             text = f"{charger.name} ({getattr(charger, 'device_id', '-')})"
-            circuit = charger.circuit
-            if circuit:
-                text += f" in {circuit.name} circuit"
-                if circuit.installation:
-                    text += f" of {circuit.installation.name} installation"
+            if charger.circuit_name:
+                text += f" in {charger.circuit_name} circuit"
+            if charger.installation:
+                text += f" of {charger.installation.name} installation"
```
sveinse, line 144: I was wondering if the user are really interested in the circuit in this flow. After we're done the user is unlikely to interact with the circuit. What do you think?

custom_components/zaptec/sensor.py
```diff
@@ -269,6 +257,13 @@ class ZapSensorEntityDescription(SensorEntityDescription):
         icon="mdi:shape-outline",
         # No state class as its not a numeric value
     ),
+    ZapSensorEntityDescription(
+        key="circuit_name",
+        translation_key="circuit_name",
+        entity_category=const.EntityCategory.DIAGNOSTIC,
+        icon="mdi:shape-outline",
+        # No state class as its not a numeric value
+    ),
```
sveinse: Do we need or want a full entity for this? I'm not sure we do. Without it the user can fetch the attribute from the "connected" entity for those few rare cases where it's needed.

custom_components/zaptec/api.py
```diff
@@ -245,19 +245,23 @@ async def build(self):
...
-            self._account.register(item["Id"], circ)
-            await circ.build()
-            circuits.append(circ)
+        chargers = []
+        for circuit_item in hierarchy["Circuits"]:
+            _LOGGER.debug("    Circuit %s", circuit_item["Id"])
+            for charger_item in circuit_item["Chargers"]:
+                _LOGGER.debug("      Charger %s", charger_item["Id"])
+                charger_item["CircuitId"] = circuit_item["Id"]
+                charger_item["CircuitName"] = circuit_item["Name"]
```
sveinse, line 260: Should we also add "CircuitMaxCurrent" as well. This is the only entity we're removing from the Circuit device, and having this in the attributes would allow users to get the information if they need to.

sveinse (changes requested):
I've read through the PR and it generally looks good. Please take a look at the specific comments I added. I've tested it on the test setup, inclusive migration from master to this PR, observing that the circuit and entities are deleted.

In addition:
* Please add a new entry for CHANGELOG.md. Add a new section on top named "Next  version" or somesuch
* There is a concept of standalone chargers which I believe is a thing of the past. See L1016-1027 in `apy.py`. I'd like to propose that we remove the `Account.stand_alone_chargers` field. In the log-entry in L1021 just warn the user that there are chargers and delete the rest of the code.

custom_components/zaptec/sensor.py
steinmn: True, removed :heavy_check_mark:

custom_components/zaptec/api.py
steinmn: Done

custom_components/zaptec/config_flow.py
steinmn: :shrug: It might be a good validation to see for new users with multiple circuits in their setup trying out HA/HACS for the first time? And all we would gain from removing it is 2 less codelines, so I'm inclined to keep it.

steinmn:
I think I've addressed all your comments + added some updates to the readme to reflect the new handling of circuits. Let me know if there is something else that needs changes.

sveinse:
I'm going to have a look now. In the meantime, would you be so kind to take a look at #193 for me, please?

custom_components/zaptec/config_flow.py
sveinse: I agree. Let's keep it.

sveinse:
Thanks a lot!

Issue #175, opened 2025-05-13
API-cal fail error 506


Amokk123:
msg :
{ domain: "zaptec", service: "stop_charging", target: object, data: object }
13.5.2025, 17:12:49[node: debug 44](http://10.10.10.10:8123/api/hassio_ingress/u64SXYtpkyFnMRsqt-wPI82fw2znIjVFhI8NWlXPnYM/#)Sum : msg.payload : number
0.4848
13.5.2025, 17:14:32[node: Turn off](http://10.10.40.109:8123/api/hassio_ingress/u64SXYtpkyFnMRsqt-xcxcxcxvdddfssdsdsdsdsdsdd)
msg : error
"HomeAssistantError: Command 'stop_charging_final' failed: Request to https://api.zaptec.com/api/chargers/xxxxx-yyyy-zzzzzz/SendCommand/506 failed after 5 retries"

This error i got each 15 min.

sveinse:
Thank you. This is another issue that what was addressed in #161.

`SendCommand` is issued when a command is issued from HA, like start and stop charging. It does not happen on its own on regular intervals. So I'm curious, what in your system is issuing stop_charging_final very frequently? It should only be issued when a session is ongoing and you want to stop the session.

Amokk123:
Now, it is not happening on its own. I have a node-red flow, thats run each 15 mintes a Day... From 20:00 it starts charge to 06:00( 10 cheapest hour between 20-06)outside this periode it sends stop charge. It has worked for aboute a year. But suddenly i got this error, without doping anything

sveinse:
Does this node-red flow issue start charge before issuing stop_charging? Are you in a charging mode when the stop charging command is issued?

FYI I use "Available Current" in the installation device to control and pause/unpause the charging. I authorize the charging and starts it (which locks the cable) and then use "available current" to 0 to temporary halt the charging. When the price is low, I raise it to the wanted amount, typically 8 to 12A (3 phase) to not exceed 10kW of the whole house. Maybe you can experiment with something similar?

Amokk123:
Okay, i will try to do something like that. I am not so good at node-red
😅😅

tir. 13. mai 2025, 20:24 skrev Svein Seldal ***@***.***>:

Amokk123:
I authenticat with a rfid card, feks 16:00. But i dont want it to start
charge the 10 cheapest hour between 20-06. This is how it have worked for a
year or so. But suddenly stopped working

[...]

PR #187, opened 2025-07-02
Do not send Pause/Resume to API if we know it will be rejected

steinmn:
The "Resume/Stop charging" buttons are disabled/grayed out when they cannot be used.

I tried doing the same for the "Charging"-switch, but it appears it just ignores the availability setting and triggers anyway. The is_command_valid check in the command-function will still block the API call if we are in a state where the command is not legal, and raising the exception triggers a popup in the UI for direct user feedback.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -676,8 +676,7 @@ class Charger(ZaptecBase):
         if command not in ZCONST.commands and command != "authorize_charge":
             if raise_value_error_if_invalid:
                 raise ValueError(f"Unknown command '{command}'")
-            else:
-                return False
+            return False
 
         if isinstance(command, int):
             # If int, look up the command name
@@ -701,12 +700,11 @@ class Charger(ZaptecBase):
 
         if valid_command:
             return True
-        elif raise_value_error_if_invalid:
+        if raise_value_error_if_invalid:
             _LOGGER.warning(msg)
             _LOGGER.debug("operation_mode: %s, final_stop_active: %s", operation_mode, final_stop_active)
             raise ValueError(msg)
-        else:
-            return False
+        return False
 
     async def set_settings(self, settings: dict[str, Any]):
         """Set settings on the charger"""
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -656,10 +656,7 @@ class Charger(ZaptecBase):
         if command == "authorize_charge":
             return await self.authorize_charge()
 
-        # Fetching the name from the ZCONST is perhaps not a good idea
-        # if Zaptec is changing them.
-        if command not in ZCONST.commands:
-            raise ValueError(f"Unknown command '{command}'")
+        self.is_command_valid(command, raise_value_error_if_invalid=True)
 
         if isinstance(command, int):
             # If int, look up the command name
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/api.py
```diff
@@ -674,6 +671,43 @@ async def command(self, command: str | int):
...
+                valid_command = False
+
+        if valid_command:
+            return True
+        elif raise_value_error_if_invalid:
+            _LOGGER.warning(msg)
+            _LOGGER.debug("operation_mode: %s, final_stop_active: %s", operation_mode, final_stop_active)
+            raise ValueError(msg)
+        else:
+            return False
```
sveinse: This was a quite extensive function. Are there easier ways we can do this?

steinmn: I'm not sure I see any simpler way to cover the functionalities I want (in a single location):
1. To block any invalid command input into the command function
2. To provide visual feedback to the user by making invalid command buttons grayed out and unclickable

1 should print a warning to the log and raise an exception if it is triggered, because the user (or a developer) has done something they're not supposed to. 2 should return true/false silently, because it will happen all the time without user interaction.

Having the flexibility to cover both cases requires some added complexity, but if we have any future additions/modifications to when commands are valid, we only have to make modifications in a single part of the code. For example, I think it would make sense to have something similar for the authorize/deauthorize buttons/commands as well, and then it's only a matter of adding another elif-case after line 700.

If you see a way to simplify the code while covering the desired functionality, I'm all ears, but I can't think of one at least. I may also be slightly environmentally damaged from being exposed to 1000+line functions in my previous job, so 36 lines doesn't look all that bad to me anymore 🤷

sveinse:
Related to #175

sveinse (commented):
On a second review, I think this is good. Please consider the nit-picks below. I'm ready to merge when you're happy.

custom_components/zaptec/api.py
```diff
@@ -674,6 +671,43 @@ async def command(self, command: str | int):
         )
         return data
 
+    def is_command_valid(self, command: str|int, raise_value_error_if_invalid=False) -> bool:
+        # Fetching the name from the ZCONST is perhaps not a good idea if Zaptec is changing them.
+        if command not in ZCONST.commands and command != "authorize_charge":
+            if raise_value_error_if_invalid:
+                raise ValueError(f"Unknown command '{command}'")
+            else:
+                return False
```
sveinse: Python nit-picks, but the else is actually not needed when there is a raise in the if. Some linters are quite obsessed by this and it saves a line. The same applies to L708-709

custom_components/zaptec/api.py
```diff
@@ -674,6 +671,43 @@ async def command(self, command: str | int):
...
+            # If int, look up the command name
+            command = ZCONST.commands.get(command)
+
+        valid_command = True
+        msg = ""
+        if command in ["resume_charging", "stop_charging_final"]:
+            # Pause/stop or resume charging are only allowed in certain states, see comments on
+            # commands 506+507 in https://api.zaptec.com/help/index.html#/Charger/Charger_SendCommand_POST
+            operation_mode = self.get("ChargerOperationMode")
+            final_stop_active = self.get("FinalStopActive")
```
sveinse, line 691: This is using the Zaptec PascalCase naming. We try to let all the internal code use the converted snake_case. This does work thou, since `get()` is converting it to the snake_case which the data is stored. However, for consistency it would be better to use the snake_case.

PS! That being said. I have come to dislike this change from the Zaptec namespace and the api internal snake_case namespace. I think everything should be using Zaptec's namespace, since this is what the documentation use. I'm working on a patch to fix this.

custom_components/zaptec/api.py
steinmn: Done. Similarly, there are some else- or elif-statements following return other places in the code that fall into the same category, but I'll leave that for now to keep the changes contained.

At some point I think it would be nice to add linting to the checks on the PRs, but that probably shouldn't be a priority right now.

custom_components/zaptec/api.py
steinmn: If it works and you're working on making changes related to this anyway, I'd rather just leave this as-is and get it merged before we get to many branches to keep track of.

sveinse:
@steinmn you can chose: Do you want this merged before or after we merge the PR for the circuits? I'm ready to merge now, and I don't think it should interfere with the other PR, but I leave it for you to chose.

sveinse:
That reminds me that I forgot to mention the CHANGELOG.md update. Oh well, have to do it when the release is made.

There are some important fixes in the PR queue, such as the reduction of bursting which should be commited before making a release.

sveinse:
@steinmn Thank you very much for your contributions 👍

Issue #188, opened 2025-07-06
Adopt to the Zaptec API fair use policy


sveinse:
With reference to the on-going discussions with Zaptec devs in #176 there is a need to make changes to this Zaptec Home Assistant integration. Currently its polling too frequently and its using mechanisms that are not documented in the API. The Zaptec integration should adhere to the [Zaptec API fair use policy](https://docs.zaptec.com/docs/api-fair-use-policy#/).

This issue is a overview of the ongoing efforts:

- [ ] #176
- [ ] #183 
- [ ] #186 
- [ ] #187

steinmn:
These are the undocumented/deprecated API-calls I've come across that don't have any PRs yet:

1. POST chargers/{self.id}/localSettings
	- Used for setting HmiBrightness and PermanentCableLock
	- Not possible to handle with current officially supported API, but these settings do seem like they would fit naturally in chargers/{self.id}/update
	- If these settings are added to chargers/{self.id}/update by Zaptec, dropping the use of localSettings becomes very simple.
2. GET chargers/{self.id}/live
	- I honestly have no idea what this does, neither on the HA side, nor on the Zaptec side. I’m not touching it unless someone explains what it does.
3. GET installation/{self.id}/messagingConnectionDetails (deprecated)
	- Should be moved to userGroups/{id}/messagingConnectionDetails
	- Where do we find "User Group ID"?
4. Authentication/Authorization
	- Used calls
		- PUT installation/{self.id} with data = {"Id": self.id, "IsRequiredAuthentication": bool}
		- POST chargers/{self.id}/authorizecharge
	- Not possible to handle with current officially supported API as far as I can see
	- authorizecharge could perhaps be handled by an added command option for chargers/{self.id}/sendCommand/{commandId}?
	- IsRequiredAuthentication could make sense to include in installation/{self.id}/update if Zaptec update the API. On the other hand, this does seem like something you only set once, so maybe we could live with having to set it from the Zaptec portal/app?

sveinse:
The use-case for authentication/authorization is two-fold:
  - Be able to have authorization via e.g. NFC tags
  - Prevent charge auto-start (delayed start)

By requiring authentication any cars plugged into the charger will not start charging immediately and the cable remains locked. This is a commonly used scenario when using HA to control when to charge. When charging is to commence, the HA integration can authorize and start the charging. This is currently the most reliable method to provide delayed start.

[...]

PR #190, opened 2025-07-11
Throttle Zaptec Cloud requests to avoid hitting rate limits (fixes #165)

### LINKED ISSUE #165: Error 429 – Too Many Requests
Hello,
I am encountering an issue with the Zaptec integration for Home Assistant. An error message appears: "Error 429 – Too Many Requests." We have 16 charging stations in the installation, and synchronization is being refused. Do you have a solution?

Thank you 

David

feliciaan:
Hello

I was running into the same issue described in #165 when using multiple Zaptec chargers — getting frequent HTTP 429 errors (“Too Many Requests”).

To fix this, I added rate limiting using the [aiolimiter](https://pypi.org/project/aiolimiter/) package, which helps queue and throttle the number of requests sent. This keeps things within Zaptec’s allowed request limits and avoids spamming retries.

This fixes it on my system with 15 Zaptec chargers connected.


sveinse:
Thank you for the contribution. This fits into the efforts to adopt to Zaptec API fair use policy as tracked in #188.

The plans are to significantly reduce the polling interval. The policy hints at 60 minutes being reasonable, while we are using 60 seconds.

May I ask how many chargers you have? And do you use/have service bus available for all of them? The polling interval shouldn't be too high if there is no service bus available.

feliciaan:
I believe the service bus is working, but is there a recommended way to verify that for sure?

The main benefit of the rate limiter is that, currently, whenever an update is triggered, requests for all chargers are sent at the same time. This creates a burst of requests on every poll, regardless of the polling interval. It’s usually not an issue with a small number of chargers, but once you go beyond 5 or 6—like in my case, with around 30 chargers—it quickly leads to hitting rate limits.

custom_components/zaptec/api.py
```diff
@@ -747,6 +748,7 @@ def __init__(
         self.is_built = False
         self._timeout = aiohttp.ClientTimeout(total=API_TIMEOUT)
         self._max_time = max_time
+        self._ratelimiter = AsyncLimiter(max_rate=10, time_period=2)
```
sveinse: Any particular reason 2 seconds is chosen here? Policy specifies 10/second

Please add suitable comments in const.py instead of using magic constants.

custom_components/zaptec/api.py
```diff
@@ -959,72 +961,73 @@ async def _refresh_token(self):
...
-                        raise log_exc(
-                            RequestDataError(f"Failed to validate data: {err}"),
-                        ) from err
-
-                    return json_result
-
-                error = RequestError(
-                        f"{method.upper()} request to {full_url} failed with status {response.status}: {response}",
-                        response.status,
+        async with self._ratelimiter:
```
sveinse: I think it would be better to place the rate limiter in the `_retry_request()` inner loop L839. This is also where the exponential back-off and jitter logic resides L872-879.

sveinse (changes requested):
When I think of it, it would have been nice if the rate limiter and the exponential backoff were a part of the same system. I wonder, does it exist something that provides both, or are we good by having them separate?

custom_components/zaptec/api.py
feliciaan: When I used with a 1 second time period, I sometimes hit the limit so I changed it two. I'll add them to the const.py.

sveinse: I can imagine that if the limit is enforced by Zaptec cloud at 10/s, its a matter of the detailed timings if we it the limit or not. So perhaps the better solution is to set down the max_rate to 9 or 8 and have the margins there instead? Have you tried that?

sveinse:
The current design in `_retry_request()` have a mechanism for exponential back-off. I'm not overly fond of having many different rate-limiting systems working all at once. I like the idea of `AsyncLimiter()` which can be used in a context. Does it exist any mechanisms that can offer both burst limiting and exponential back-off on retries? If not, should this be isolated out to a separate function which includes both?

custom_components/zaptec/api.py
feliciaan: I'll now test it with 10 requests per second, but moved to the place you suggested.

sveinse (approved):
I think this looks good. I've tested it and the throttling seems to work fine. I'd like to change a little bit on how the exponential back-off is implemented, but we'll do that in a follow-up PR. Approved.

PR #191, opened 2025-07-14
Support for reconfigure and accompanying improvements

### LINKED ISSUE #76: Add support to reconfigure zaptec
Currently it is not possible to reconfigure zaptec. The only way to change username and password is to delete the integration and set it up anew. It should be easier for the user to reconfigure this.

(From sveinse/zaptec#9)

sveinse:
This PR started with the intention to fix #76:  Reconfigure the Zaptec integration. However, once pulling the threads in the code revealed a few things that had to be fixed for this to work. When reconfiguring an integration, it needs to shut down the old to start a new, and the current design didn't do that very well. In particular the stream methods didn't clean up properly and were failing miserably on reconfigure.

Next the setup weren't implemented in a good way, as any login/password issues should be handled earlier so it had to be improved. We needed a separate login method instead of just refreshing the token when doing the first access.

There are a few method being changed from this PR. However the `api.py` file exists to serve the integration, so I think its ok to do these changes.

* Refactor of configuration flows
  * Add support for reconfigure config flow
  * Add support for reauthenticate config flow
  * Refactor user setup config flow
* Refactor stream methods
  * Rename Installation._stream() to stream_main()
  * Move Account.update() to Installation.stream_update()
  * Make sure only one stream can run
  * Implement Installation.stream_close()
* Refactor API request methods
  * Remove Account.check_login()
  * Add Account.login()
  * Rename Account._retry_request() to _request_worker()
* Improve integration setup
  * Move Account initalization to early setup and fail there
  * Handle streams from the coordinator
  * Move first-time-setup to ZaptecUpdateCoordinator._first_time_setup()

This PR is the first in a series to move towards a more Zaptec API friendly design.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -1,4 +1,5 @@
 """Main API for Zaptec."""
+
 from __future__ import annotations
 
 import asyncio
@@ -6,18 +7,26 @@ import json
 import logging
 import random
 from abc import ABC, abstractmethod
-from collections import UserDict
 from collections.abc import Iterable
-from concurrent.futures import CancelledError
 from contextlib import aclosing
-from typing import Any, AsyncGenerator, Callable, Protocol, cast
+from typing import Any, AsyncGenerator, Callable, Protocol
 
 import aiohttp
 import pydantic
 
 from .const import (
-    API_RETRIES, API_RETRY_FACTOR, API_RETRY_JITTER, API_RETRY_MAXTIME,
-    API_TIMEOUT, API_URL, MISSING, TOKEN_URL, TRUTHY, CHARGER_EXCLUDES)
+    API_RETRIES,
+    API_RETRY_FACTOR,
+    API_RETRY_INIT_DELAY,
+    API_RETRY_JITTER,
+    API_RETRY_MAXTIME,
+    API_TIMEOUT,
+    API_URL,
+    CHARGER_EXCLUDES,
+    MISSING,
+    TOKEN_URL,
+    TRUTHY,
+)
 from .misc import mc_nbfx_decoder, to_under
 from .validate import validate
 from .zconst import ZConst
@@ -37,6 +46,7 @@ _LOGGER = logging.getLogger(__name__)
 
 # Set to True to debug log all API calls
 DEBUG_API_CALLS = False
+DEBUG_API_DATA = False
 
 # Set to True to debug log all API errors
 # Setting this to False because the error messages are very verbose and will
@@ -230,6 +240,7 @@ class Installation(ZaptecBase):
 
         self._stream_task = None
         self._stream_receiver = None
+        self._stream_running = False
 
     async def build(self):
         """Build the installation object hierarchy."""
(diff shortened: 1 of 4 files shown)


custom_components/zaptec/api.py
```diff
@@ -380,33 +409,59 @@ async def _stream(self, cb=None, ssl_context=None):
...
+            return
+
+        if (charger := self._account.map.get(charger_id)) is None:
+            _LOGGER.warning("Got update for unknown charger id %s", charger_id)
+            return
+
+        charger = self.map.get(charger_id)
+        if charger is None:
+            _LOGGER.warning("Got update for unknown charger id %s", charger_id)
+            return
```
steinmn: This looks like a copypasta gone wrong, I think the first charger assignment is the correct one, and the second is an artifact from before the function was moved from the account object?

sveinse: Good catch. It is right, but I changed the text a little bit to be a little more informative of the error.

steinmn: But the installation class doesn't have a map attribute, that is in the account class:
```
2025-07-15 13:14:38.060 ERROR (MainThread) [custom_components.zaptec.api] Couldn't process stream message: 'map'
  File "/workspaces/zaptec/custom_components/zaptec/api.py", line 158, in __getattr__
    return self._attrs[to_under(key)]
           ~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'map'

The above exception was the direct cause of the following exception:

  File "/workspaces/zaptec/custom_components/zaptec/api.py", line 402, in stream_main
    self.stream_update(json_result)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/workspaces/zaptec/custom_components/zaptec/api.py", line 467, in stream_update
    charger = self.map.get(charger_id)
              ^^^^^^^^
  File "/workspaces/zaptec/custom_components/zaptec/api.py", line 160, in __getattr__
    raise AttributeError(exc) from exc
AttributeError: 'map'
```

custom_components/zaptec/api.py
```diff
@@ -407,31 +420,58 @@ async def _stream(self, cb=None, ssl_context=None):
...
+
+        if (charger := self._account.map.get(charger_id)) is None:
+            _LOGGER.warning("Got update for unknown charger id %s", charger_id)
+            return
+
+        charger = self.map.get(charger_id)
+        if charger is None:
+            _LOGGER.warning("Got update for unregistered charger, id %s", charger_id)
+            return
```
```suggestion
        if (charger := self._account.map.get(charger_id)) is None:
            _LOGGER.warning("Got update for unregistered charger, id %s", charger_id)
            return
```

custom_components/zaptec/api.py
sveinse: Oh, right. How did that miss my testing? I'll fix this.

custom_components/zaptec/api.py
sveinse: New version pushed

steinmn (approved):
LGTM

PR #193, opened 2025-07-14
Adopt the exponential back-off to the rate limiter

sveinse:
As a follow-up of #190, adopt the exponential back-off mechanism to account for the time the rate limiter might delay the retries.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -5,6 +5,7 @@ import asyncio
 import json
 import logging
 import random
+import time
 from abc import ABC, abstractmethod
 from collections import UserDict
 from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/api.py
```diff
@@ -908,9 +912,14 @@ def log_exc(exc: Exception) -> Exception:
...
-                if DEBUG_API_CALLS:
-                    _LOGGER.debug("Sleeping for %s seconds", delay)
-                await asyncio.sleep(delay)
+
+                # If the sleep time is negative, it means the request took
+                # longer than the wanted delay, so we don't need to sleep.
+                sleep_delay = delay - time.perf_counter() + start_time
+                if sleep_delay > 0:
+                    if DEBUG_API_CALLS:
+                        _LOGGER.debug("Sleeping for %1.1f seconds", sleep_delay)
```
steinmn, line 921: With 0.01 as the initial delay, won't this just print out 0.0, or am I just too tired? Shouldn't we use `%.2f` instead? (the minimum width of 1 is enforced by the number of decimals anyway)

sveinse: Good catch. I'll put this into the next update

PR #198, opened 2025-07-16
Main API cleanup. Breaking changes

sveinse:
* class Zaptec
  * Make class Zaptec an async context in order to close the web session
  * Make the class a `Mapping` of (Zaptec) objects with dict-like interface.
  * Mapping changes must go via `register` and `unregister`.
  * `register` will fail on re-registeringing an existing object
  * Add `.objects()`, `.installations` and `chargers` as methods to fetch subset of data from the class
  * `.build()` can be called multiple times. If run again it will update the existing objects
  * Stop exposing `.map`
  * Remove `get_circuits_id()`. Its too specialized and is handled directly where its used.
  * `.get_chargers()` is not the property `.chargers`
  * `.installations` is now a property and not a variable
  * Add warnings on removal of or standalone chargers. We don't handle this yet.

* class ZaptecBase
  * Make class ZaptecBase a `Mapping` with a RO dict-like interface for parameters
  * Removed `.__getattr__()` so `obj.something` doesn't work. However `.id` and `.name` is explicitly kept.

* class Installation
  * Update `.build()` to support re-runs without creating new objects

* class Charger
  * Remove unused `.live()` method
  * Remove empty `.build()` method

* General
  * Cleanup and reorganization
  * Adopt the HA integration to the API changes

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -261,9 +261,8 @@ class Installation(ZaptecBase):
         for circuit in hierarchy["Circuits"]:
             _LOGGER.debug("    Circuit %s", circuit["Id"])
             for charger_item in circuit["Chargers"]:
-
                 # Inject additional attributes
-                charger_item["InstallationId"] = self.id  # So the relationship is ready at build
+                charger_item["InstallationId"] = self.id
                 charger_item["CircuitId"] = circuit["Id"]
                 charger_item["CircuitName"] = circuit["Name"]
                 charger_item["CircuitMaxCurrent"] = circuit["MaxCurrent"]
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -144,9 +144,6 @@ class ZaptecBase(Mapping[str, TValue]):
     # =======================================================================
     #   UPDATE METHODS
 
-    async def build(self) -> None:
-        """Build the object."""
-
     async def state(self) -> None:
         """Update the state of the object."""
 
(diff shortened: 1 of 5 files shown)


custom_components/zaptec/api.py
```diff
@@ -1106,50 +1157,54 @@ async def build(self):
...
+        new_chargers = {d["Id"] for d in chargers["Data"]}
+        have_chargers = {c.id for c in self.chargers}
+        if missing_chargers := (have_chargers - new_chargers):
+            _LOGGER.warning(
+                "These chargers are no longer available: %s", missing_chargers
+            )
+            _LOGGER.warning("To remove them, please restart the integration.")
+        if extra_chargers := (new_chargers - have_chargers):
+            _LOGGER.warning(
+                "These standalone chargers will not added: %s", extra_chargers
```
```suggestion
                "These standalone chargers will not be added: %s", extra_chargers
```

custom_components/zaptec/api.py
```diff
@@ -256,20 +260,29 @@ async def build(self):
...
+                if charger_item["Id"] in self.zaptec:
+                    _LOGGER.debug("      Charger %s  (existing)", charger_item["Id"])
+                    charger: Charger = self.zaptec[charger_item["Id"]]
+                    charger.set_attributes(charger_item)
+                else:
+                    _LOGGER.debug("      Charger %s", charger_item["Id"])
+                    charger = Charger(charger_item, self.zaptec, installation=self)
+                    self.zaptec.register(charger_item["Id"], charger)
+
+                await charger.build()
```
steinmn: Do we need this? charger.build() still does nothing, right?

custom_components/zaptec/api.py
sveinse: Fixed. Good catch.

custom_components/zaptec/api.py
sveinse: I first argued that it should be there for consistency. But then I was thinking that if we need something to happen in `Charger.build()` we have to run a PR for it, so why not also update this part then? So I removed it.

Issue #189, opened 2025-07-06
Create more reliable update from Zaptec on changes


sveinse:
Based on finding in https://github.com/custom-components/zaptec/pull/183#issuecomment-3041289723 to solve #176, it shows that data is slow to get update from the Zaptec cloud, and more importantly, it can return data which is outdated. This is a challenge for the user as the user will either see data which is slow to update or values jumping back and forth between `NEW -> OLD -> NEW`.

Below I have some example tracking the changes of `ChargerMaxCurrent` to highlight the issues. This issue is not limited to this single parameter only, but is a general challenge of how to reliably getting updates from Zaptec cloud when changes are made.

The objective of this issue to discuss how we can better adopt the API interface logic in the HA integration to the current behavior of the cloud.

## Case 1

In this example the max charger current is first set by the user. A service bus push is received, but it does _not_ contain an update of the `ChargerMaxCurrent`. This parameter is not updated until the polling of the charger state is done.

If we implement very infrequent polling (like 1 hour) and does not do forced polls after a command, this means that it will take up to the length of a poll interval to properly update the values of changed parameters.

```sh
# Polling state
18:09.783 [zaptec.api] Polling state for Charger[f40861] (Sol)
18:09.877 [zaptec.api]     ChargerMaxCurrent = 32.000 (2025-07-06T12:14:14.543)

# Setting max current
18:34.470 [zaptec.number] Setting ZaptecSettingNumber.charger_max_current to <float> 15.0   (in f40861)
18:34.470 [zaptec.api] Settings {'maxChargeCurrent': 15.0}

# Getting notified from service bus
# NOTE That the service bus does not contain `ChargerMaxCurrent` events
18:35.235 [zaptec.api] ---   Subscription: {'DeviceId': 'xxx', 'DeviceType': 4, 'ChargerId': 'xxx', 'StateId': '511 (ChargerMinCurrent)', 'Timestamp': '2025-07-06T11:35:56.453', 'ValueAsString': '5.000'}
18:35.256 [zaptec.api] ---   Subscription: {'DeviceId': 'xxx', 'DeviceType': 4, 'ChargerId': 'xxx', 'StateId': '711 (IsEnabled)', 'Timestamp': '2023-07-12T13:29:38.207', 'ValueAsString': '1'}

# Forced polling to which gets the correct state value
18:35.800 [zaptec.api] Polling state for Charger[f40861] (Sol)
[...]

PR #201, opened 2025-07-18
Refactor Zaptec integration to improve polling and command handling

sveinse:
- New feature: When HA initiates a change or command, the integration will poll the device 2, 5 and 14 seconds after to fetch updated data quickly.
- Introduced polling methods for charger and installation updates.
- Replaced direct state updates with trigger_poll calls in various components.
- Updated command handling in services and entities to use a unified command method.
- Removed direct commands in favor of the command method

> [!NOTE]
> This PR contains a known issue: It contains a new feature which issue a sequence of polls immediately after any HA imitated command or change. If HA makes very frequent changes (<20 secs), there might be many updates running in parallel for the same resource which is not good. This is planned to be fixed in a later PR.

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -55,6 +55,8 @@ from .const import (
     MANUFACTURER,
     MISSING,
     REQUEST_REFRESH_DELAY,
+    ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS,
+    ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS,
 )
 from .services import async_setup_services, async_unload_services
 
@@ -133,7 +135,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     # Make a set of the circuit ids from zaptec to check for deprecated Circuit-devices
     circuit_ids = {
-        cid for c in coordinator.zaptec.chargers if (cid := c.get("circuit_id", ""))
+        cid for c in coordinator.zaptec.chargers if (cid := c.get("CircuitId"))
     }
 
     # Clean up unused device entries with no entities
@@ -193,6 +195,7 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
         self.zaptec: Zaptec = zaptec
         self.streams: list[tuple[asyncio.Task, Installation]] = []
         self.entity_maps: dict[str, dict[str, ZaptecBaseEntity]] = {}
+        self.updateable_objects: set[str] = set()
 
         super().__init__(
             hass,
@@ -277,11 +280,13 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
         if self.config_entry.data.get(CONF_MANUAL_SELECT, False):
             chargers = self.config_entry.data.get(CONF_CHARGERS)
 
+        all_objects = set(self.zaptec)
+        self.updateable_objects = all_objects
+
         # Selected chargers to add
         if chargers is not None:
             _LOGGER.debug("Configured chargers: %s", chargers)
             want = set(chargers)
-            all_objects = set(self.zaptec)
 
             # Log if there are any objects listed not found in Zaptec
             not_present = want - all_objects
@@ -300,11 +305,8 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
             if not keep:
                 _LOGGER.error("No zaptec objects will be added")
 
-            # Unregister all discovered objects not in the keep list
-            for objid in all_objects:
-                if objid not in keep:
-                    _LOGGER.debug("Unregistering: %s", objid)
-                    self.zaptec.unregister(objid)
+            # These objects will be updated by the coordinator
+            self.updateable_objects = want
 
         # Setup the stream subscription
         self.create_streams()
@@ -323,13 +325,55 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
 
                 # Fetch updates
                 _LOGGER.debug("Polling from Zaptec")
-                await self.zaptec.update_states()
+                await self.zaptec.poll(
+                    self.updateable_objects, state=True, info=True, firmware=True
+                )
 
         except ZaptecApiError as err:
             _LOGGER.exception(
                 "Fetching data failed: %s: %s", type(err).__qualname__, err
             )
             raise UpdateFailed(err) from err
+
+    async def _trigger_poll(self, obj: ZaptecBase) -> None:
+        """Trigger a poll update sequence for the given object or all objects.
+
+        This sequence is useful to ensure that the state is fully synced after a
+        HA initiated update.
+        """
+
+        what = {obj.id}
+        if isinstance(obj, Installation):
+            delay_list = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
+            kw = {"state": True, "info": True}
+        else:
+            delay_list = ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS
+            kw = {"state": True}
+
+        for i, delay in enumerate(delay_list, start=1):
+            await asyncio.sleep(delay)
+            _LOGGER.debug(
+                "Triggering poll %s of %s after %s seconds. %s",
+                i,
+                obj.qual_id,
+                delay,
+                kw,
+            )
+            await self.zaptec.poll(what, **kw)
+            self.async_update_listeners()
+
+    async def trigger_poll(self, obj: ZaptecBase) -> None:
+        """Trigger a poll update sequence."""
+
+        # FIXME: The current imeplementation will create a new background task
+        # no matter if a task is already running. If they are updating the same
+        # object, this can cause too many updates for the same object.
+        # A single task per device will be needed.
+        self.config_entry.async_create_background_task(
+            self.hass,
+            self._trigger_poll(obj),
+            f"Zaptec Poll Update for {obj.qual_id if obj else 'all'}",
+        )
 
 
 class ZaptecBaseEntity(CoordinatorEntity[ZaptecUpdateCoordinator]):
@@ -533,3 +577,7 @@ class ZaptecBaseEntity(CoordinatorEntity[ZaptecUpdateCoordinator]):
                 _LOGGER.error("Unknown zaptec object type: %s", type(obj).__qualname__)
 
         return entities
+
+    async def trigger_poll(self) -> None:
+        """Trigger a poll for this entity."""
+        await self.coordinator.trigger_poll(self.zaptec_obj)
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -344,13 +344,16 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
 
         what = {obj.id}
         if isinstance(obj, Installation):
-            delay_list = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
+            delays = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
             kw = {"state": True, "info": True}
         else:
-            delay_list = ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS
+            delays = ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS
             kw = {"state": True}
 
-        for i, delay in enumerate(delay_list, start=1):
+        # Calculcate the deltas for the delays. E.g. [2, 5, 10] -> [2, 3, 5]
+        deltas = [b - a for a, b in zip([0] + delays[:-1], delays)]
+
+        for i, delay in enumerate(deltas, start=1):
             await asyncio.sleep(delay)
             _LOGGER.debug(
                 "Triggering poll %s of %s after %s seconds. %s",
@@ -372,7 +375,7 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
         self.config_entry.async_create_background_task(
             self.hass,
             self._trigger_poll(obj),
-            f"Zaptec Poll Update for {obj.qual_id if obj else 'all'}",
+            f"Zaptec Poll Update for {obj.qual_id}",
         )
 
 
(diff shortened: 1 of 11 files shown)


custom_components/zaptec/__init__.py
```diff
@@ -323,14 +325,56 @@ async def _async_update_data(self) -> None:
...
+                )
 
         except ZaptecApiError as err:
             _LOGGER.exception(
                 "Fetching data failed: %s: %s", type(err).__qualname__, err
             )
             raise UpdateFailed(err) from err
 
+    async def _trigger_poll(self, obj: ZaptecBase) -> None:
+        """Trigger a poll update sequence for the given object or all objects.
```
steinmn: For this to work for all objects, doesn't obj need to be None? ie `obj: ZaptecBase | None` And doesn't that also break for obj.id and obj.qualid?

custom_components/zaptec/api.py
```diff
@@ -728,26 +734,6 @@ async def set_settings(self, settings: dict[str, Any]):
...
-
-    async def deauthorize_and_stop(self):
-        """Deauthorize the charger and stop it."""
-        return await self.command("deauthorize_and_stop")
-
-    async def restart_charger(self):
-        """Restart the charger."""
-        return await self.command("restart_charger")
-
-    async def upgrade_firmware(self):
```
steinmn, line 747: Removing this and relying solely on string inputs does seem like it could be prone to typo errors, should we consider adding an enum or a  Commands-class with constants with the supported options?

custom_components/zaptec/const.py
```diff
@@ -37,6 +37,12 @@
 API_RATELIMIT_MAX_REQUEST_RATE = 10
 """Maximum number of requests allowed per API rate limit period."""
 
+ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS = [2, 5, 7]
```
steinmn: This corresponds to 2, 7 and 14 seconds after trigger, as opposed to the 2, 5 and 14 in the Pr description. Should perhaps also include a comment with the cumulative values, since that was used in the description of the PR.

custom_components/zaptec/const.py
```diff
@@ -37,6 +37,12 @@
...
+""" Sequence of delays in seconds for state updates to a charger after a change."""
+
+ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS = [2, 5]
```
steinmn: Maybe worth a comment on why we only do 2 delays for the installations?

custom_components/zaptec/__init__.py
```diff
@@ -323,14 +325,56 @@ async def _async_update_data(self) -> None:
...
+        for i, delay in enumerate(delay_list, start=1):
+            await asyncio.sleep(delay)
+            _LOGGER.debug(
+                "Triggering poll %s of %s after %s seconds. %s",
+                i,
+                obj.qual_id,
+                delay,
+                kw,
+            )
+            await self.zaptec.poll(what, **kw)
```
steinmn, line 365: What if this takes a long time/times out, do we still want to add the same delays?
Example: first poll times out (10 seconds?). Second poll succeeds. We are now 2+10+5=17 seconds since polling was triggered, do we really wait for 7 seconds and trigger a 3rd time?

sveinse:
> Doesn't the Zaptec API cache values for 15 seconds or did I misunderstand the docs ( https://docs.zaptec.com/docs/api-fair-use-policy#2-api-usage-and-rate-limiting)? Would we really gain anything by polling multiple times within a 15 second span?

I am disucussing a lot with Zaptec about this. Notice what is said in https://github.com/custom-components/zaptec/issues/176#issuecomment-3062637738

"There is no caching like mentioned https://github.com/custom-components/zaptec/issues/189#issuecomment-3042711866, this is outdated info (fixed now). So, if value in states in still old one it only means that charger hasn't reported it back yet."

custom_components/zaptec/api.py
sveinse: Hmm. I'm not so sure. I don't really want to duplicate the command enum already available from Zaptec constants. DRY.

It does open up the possibility that our code breaks if Zaptec renames their commands, but then we're back at needing to use numerical IDs for identification of command and keeping our own record of commands.

It _is_ more explicit having dedicated functions. I do agree that finding the exact list of commands Zaptec supports take some doing. That's why I updated the comment for the command function.

custom_components/zaptec/__init__.py
sveinse: Yes, the current setup is not ideal. Normally the coordinator takes care of this.

My plan is to refine this implementation in coming changes. To implement the device state dependent poll-interval, we need to implement a coordinator (with its own intervals) per charger. This poll sequencing and the overall poll-interval thingy must work together so we don't get a stampede of requests.

Especially if timeouts gets involved. Coordinator takes care of that, so there is no need to reinvent the wheel.

custom_components/zaptec/const.py
sveinse: Heh. Is "gut feeling" comment worthy?

custom_components/zaptec/const.py
sveinse: Would it be more natural to use the absolute times since the command, e.g. [2, 7, 14], instead of deltas? Because it would be simple enough to calculate the required delays on the fly. That is perhaps less ambiguous? What do you think?

custom_components/zaptec/api.py
steinmn: I see your point. Then again, in https://docs.zaptec.com/reference/api_chargers_id_sendcommand_commandid_post only 5 commands are listed as supported, so does that mean that we cannot rely on all the commandids returned by api/constants, or is api/constants considered self-documenting?

custom_components/zaptec/__init__.py
sveinse: `_trigger_poll()` is never called without `obj` so I don't really understand the question?

Have you found anywhere where `ZaptecUpdateCoordinator.trigger_poll()` is called without an object?

steinmn: Then what is the input given to _trigger_poll that polls all objects (since the docstring says it can do that)?

custom_components/zaptec/const.py
steinmn: Yes, I think that would be preferable

custom_components/zaptec/__init__.py
steinmn:
```
self.config_entry.async_create_background_task(
            self.hass,
            self._trigger_poll(obj),
            f"Zaptec Poll Update for {obj.qual_id if obj else 'all'}",
        )
```
in trigger_poll, the `if obj` does seem to imply that obj can be None/empty, but I cant see that trigger_poll is called without an object anywhere, no.

custom_components/zaptec/api.py
sveinse: I'll think about it. -- It does not help that we have this "ResumeCharging" -> "resume_charging" obfuscation. That will have to go at some point. It is very confusing - even for me with many, many hours deep dived into the code.

custom_components/zaptec/__init__.py
sveinse: Oh, that `if` is a remnant of old code. Good catch. I'll fix it.

custom_components/zaptec/api.py
sveinse: I have an idea, which I'd like to run by you to see if its good:

The command can be anything, and the list provided by Zaptec is very long. However, they publicly only recommend a few. So what if we do this at type annotation instead:

```py
type CommandType = Literal[
    "RestartCharger",
    "UpgradeFirmware",
    "StopChargingFinal",
    "ResumeCharging",
    "DeauthorizeAndStop",
]
"""Approved API commands."""
```

Then it can be listed in the editor:

image

What do you think about this? It doesn't change anything functionally, it only aids the developer using this.

custom_components/zaptec/__init__.py
sveinse: Fixed

custom_components/zaptec/const.py
sveinse: Fixed. Please review if the implementation and comments looks better.

custom_components/zaptec/__init__.py
steinmn: Is there an input that polls all objects (in this function)? If not:
```suggestion
        """Trigger a poll update sequence for the given object.
```

custom_components/zaptec/api.py
```diff
@@ -616,52 +641,44 @@ async def state(self):
...
+        # Look up the command and its command id
         if isinstance(command, int):
             # If int, look up the command name
             cmdid = command
-            command = ZCONST.commands.get(command)
+            command = ZCONST.commands.get(cmdid)
         else:
-            cmdid = ZCONST.commands.get(command)
+            # Support using the CommandName as a string
+            cmdid = ZCONST.commands.get(to_under(command))
```
steinmn, line 674:
```suggestion
            command = to_under(command)
            cmdid = ZCONST.commands.get(command)
```
Otherwise we need to cover both PascalCase and snake_case in is_command_valid (or run to_under on the input to is_command_valid if that is preferable).

sveinse: Yes, you're right

sveinse: But I think the code is correct as it is. By the time `is_command_value()` is called in `command()` command is always snake_case. So for now, `is_command_valid()` only operates on snake_casse. Entities use snake_case too, so the `is_command_valid()` call from the button class is also valid.

I think cleaning up the usage of pascal and snake case will be a later PR. But we need it soon, because it makes it more complicated.

custom_components/zaptec/api.py
sveinse: PS! I haven't committed it yet. Do you want me to commit it for you to test it?

custom_components/zaptec/__init__.py
```diff
@@ -323,14 +325,59 @@ async def _async_update_data(self) -> None:
...
+            delays = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
+            kw = {"state": True, "info": True}
+        else:
+            delays = ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS
+            kw = {"state": True}
+
+        # Calculcate the deltas for the delays. E.g. [2, 5, 10] -> [2, 3, 5]
+        deltas = [b - a for a, b in zip([0] + delays[:-1], delays)]
+
+        for i, delay in enumerate(deltas, start=1):
```
steinmn: At first glance, this looked like a bug since we have a `delays` list that shouldn't be used directly for the sleep, while we are using `delay` here. (deltas and delays being the same length with 4/6 letters in common certainly doesn't help)

While I realize it is correct as written, wouldn't it be better to use `delta` as the variable in the loop?

custom_components/zaptec/api.py
steinmn: Yes, would be nice to test.

Personally I think I would have gone the enum-route and relied more on comparing them to commandids to the enum int values, but this is probably the simpler route with everything being a little in flux right now.

custom_components/zaptec/api.py
steinmn: Yes, I think all current uses are snake_case, and cleaning up the mix of Pascal and snake would definitely make it easier going forward.

custom_components/zaptec/const.py
steinmn: Looks good now 👍  (minor nitpick/question on loop variable name in separate comment)

custom_components/zaptec/api.py
steinmn: Did a quick test just copypasting your code above in:
image
I think this is ok

steinmn: Do we include `AuthorizeCharge` in the list? It's not really approved, but it is one we rely on in the integration :shrug:

custom_components/zaptec/__init__.py
sveinse: The straight forward approach is more verbose:

```py
prev = 0
for i, delay in enumerate(delays, start=1):
    delta = delay - prev
    prev = delay
```

I guess this boils down to preference and readability.

steinmn: The onliner definition of deltas is fine by me, I'm asking if
```suggestion
        for i, delta in enumerate(deltas, start=1):
```
(and using delta in the rest of the for loop as well) is easier to read? For me at least, `for delta in deltas` feels more natural, especially when there is already a `delays` list within the same code block.

custom_components/zaptec/api.py
sveinse: Yes, let's include it (with a comment).

> Personally I think I would have gone the enum-route and relied more on comparing them to commandids to the enum int values, but this is probably the simpler route with everything being a little in flux right now.

What kind of enum flavor do you mean here? `class Enum` type? Won't that require more code to extract the value from it?

custom_components/zaptec/__init__.py
sveinse: yes, I agree.

You have a keen eye for details like this that evidently pass my eyes. Thanks. 👍

custom_components/zaptec/api.py
steinmn: I was thinking something like 
image
and then comparing commandids against that
image
instead of comparing strings in order to be a bit more robust against zaptec potentially changing naming.

I guess you could also just drop the Enum-inheritance and have a basic class of named constants so that you can skip the int() casting.
image

But the more I think about it, the messier it becomes, so CommandType + a cleanup of Pascal/snake is probably better

sveinse: I see. I think agree that this gets messy. We're relying on the stability of the Zaptec provided ID and name mapping. And I think you're on to something that the numerical ID is probably more stable than the name from them. So its not a bad idea to compare against ID rather than names. The principal question is: Will our maintenance of name <-> id mapping more stable than Zaptec'?

Anyhow, I think I'll stick with the middle ground using CommandType for now. I have much more in store for changes to the HA parts of the integration which need to focus on.

sveinse: FIx committed and pushed. Please see how it turned out. I added some additional docstrings in the class ZConst to further assist.

custom_components/zaptec/__init__.py
sveinse: Fixed

steinmn (approved):
LGTM

Issue #202, opened 2025-07-18
Zaptec polling intervals


sveinse:
To reduce the polling foot-print on Zaptec cloud, there is a need to implement context dependent poll-intervals. 

This is the current proposal based on conversation with Zaptec. See https://github.com/custom-components/zaptec/issues/176#issuecomment-3080807935

| State | Endpoints (`GET`) | Poll interval |
|---|---|---|
| Chargers: Charging | `chargers/{id}/state` | 60 second |
| Chargers: Disconnected/idle | `chargers/{id}/state` | 10 minute |
| Chargers: Info | `charger/{id}` | 60 minute |
| Installation: Info | `installation/{id}` | 10 minute |
| Firmware updates | `chargerFirmware/installation/{id}` | 1 day |
| Structure | `installation`, `installation/{id}/hierarchy` | ~~1 day~~ Only at startup |
| Charger commands | `chargers/{id}/state` | 2, 7, 15 seconds (1) |
| Installation commands | `installation/{id}` | 2, 7 seconds (2) |

* (1) "Commands" are any action from HA that results in `POST` or `PUT` sent to Zaptec. This will happen only once. The resulting sequence of polling at intervals is using `GET` state to ensure the local data is updated sufficiently fast for UX and are able to handle latencies on the cloud side. The exact delay values can be fine tuned later.
* (2) Installations doesn't have state, so the only end-point to use for polling is `installation/{id}´

steinmn:
Should (some) installation commands also trigger polls of associated chargers? For instance, updating available_current for an installation will most likely change the charge power of the chargers in it.

PR #203, opened 2025-07-19
Refactor Zaptec integration for prepping to adopt to API policy

sveinse:
This PR introduces a lot of refactoring to do several things
- Modernize the way the integration is set up - move toward HA coding ideals, keeping the door open to future inclusion into HA.
- Introduce a ZaptecManager that will later allow us to store multiple coordinator. This will be required to do the elaborate polling setup that Zaptec wants, see #202 

> [!WARNING]
> Currently this is WIP/draft and rather untested, but adding it here so its possible to do early reviews on it. All input helps.

From commit message:
- Introduced ZaptecManager to handle instance data and coordination of updates.
- It is preparations for running multiple coordinators
- Updated async_setup_entry to utilize the new manager
- Refactored entity creation methods to be part of the manager
- Adjusted all platform files (binary_sensor, button, lock, number, sensor, switch, update) to align with the new structure.
- Cleaned up unnecessary imports and comments, enhancing code readability.
- Removed YAML import step from config flow as it is no longer needed.

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Iterable
 from copy import copy
+from dataclasses import dataclass
 from datetime import timedelta
 import logging
 from typing import Any
@@ -27,7 +28,6 @@ from homeassistant.helpers import device_registry as dr, entity_registry as er
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
 from homeassistant.helpers.debounce import Debouncer
 from homeassistant.helpers.entity import DeviceInfo, EntityDescription
-from homeassistant.helpers.typing import ConfigType
 from homeassistant.helpers.update_coordinator import (
     CoordinatorEntity,
     DataUpdateCoordinator,
@@ -46,7 +46,6 @@ from .api import (
     ZaptecBase,
 )
 from .const import (
-    API_TIMEOUT,
     CONF_CHARGERS,
     CONF_MANUAL_SELECT,
     CONF_PREFIX,
@@ -62,25 +61,34 @@ from .services import async_setup_services, async_unload_services
 
 _LOGGER = logging.getLogger(__name__)
 
+type ZaptecConfigEntry = ConfigEntry[ZaptecManager]
+
 PLATFORMS = [
     Platform.BINARY_SENSOR,
     Platform.BUTTON,
-    # Platform.DEVICE_TRACKER,
     Platform.LOCK,
-    # Platform.NOTIFY,
     Platform.NUMBER,
-    # Platform.SELECT,
     Platform.SENSOR,
     Platform.SWITCH,
     Platform.UPDATE,
 ]
 
 
-async def async_setup(hass: HomeAssistant, config: ConfigType) -> bool:
-    """Set up integration."""
-    if DOMAIN in hass.data:
-        _LOGGER.info("Delete zaptec from your yaml")
-    return True
+@dataclass(frozen=True, kw_only=True)
+class ZaptecEntityDescription(EntityDescription):
+    """Class describing Zaptec binary sensor entities."""
+
+    cls: type[ZaptecBaseEntity]
+
+
+@dataclass(frozen=True, kw_only=True)
+class ZaptecUpdateOptions:
+    """Options for the Zaptec update coordinator."""
+
+    name: str
+    update_interval: int
+    tracked_devices: set[str]
+    poll_args: dict[str, bool]
 
 
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
@@ -93,13 +101,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     _LOGGER.debug("Setting up entry %s: %s", entry.entry_id, redacted_data)
 
-    # Create the Zaptec account object and log in
+    configured_chargers = None
+    if entry.data.get(CONF_MANUAL_SELECT, False):
+        configured_chargers = entry.data.get(CONF_CHARGERS)
+
+    prefix = entry.data.get(CONF_PREFIX, "").rstrip()
+    if prefix:
+        prefix = prefix + " "
+
+    scan_interval = entry.data.get(CONF_SCAN_INTERVAL, DEFAULT_SCAN_INTERVAL)
+
+    # Create the Zaptec object
     zaptec = Zaptec(
         entry.data[CONF_USERNAME],
         entry.data[CONF_PASSWORD],
         client=async_get_clientsession(hass),
-        max_time=entry.data.get(CONF_SCAN_INTERVAL, DEFAULT_SCAN_INTERVAL),
+        max_time=scan_interval,
     )
+
+    # Login to the Zaptec account
     try:
         await zaptec.login()
     except AuthenticationError as err:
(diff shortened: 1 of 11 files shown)


custom_components/zaptec/__init__.py
```diff
@@ -62,25 +61,34 @@
...
-async def async_setup(hass: HomeAssistant, config: ConfigType) -> bool:
-    """Set up integration."""
-    if DOMAIN in hass.data:
-        _LOGGER.info("Delete zaptec from your yaml")
-    return True
+@dataclass(frozen=True, kw_only=True)
+class ZaptecEntityDescription(EntityDescription):
+    """Class describing Zaptec binary sensor entities."""
+
+    cls: type[ZaptecBaseEntity]
```
joostlek: why is the class part of it?

joostlek: You can move this shared thing to entity.py

custom_components/zaptec/__init__.py
```diff
@@ -185,50 +221,124 @@ async def async_reload_entry(hass: HomeAssistant, entry: ConfigEntry) -> None:
...
+    """Set of tracked device that will be used by the integration."""
+
+    name_prefix: str
+    """Name prefix to use for the entities."""
+
+    coordinator: ZaptecUpdateCoordinator
+    """The coordinator for the Zaptec account."""
+
+    streams: list[tuple[asyncio.Task, Installation]] = []
+    """List of active streams for the installations."""
```
joostlek: We also regularly check for new devices right? Isn't this in essence also a coordinator?

custom_components/zaptec/__init__.py
```diff
@@ -185,50 +221,124 @@ async def async_reload_entry(hass: HomeAssistant, entry: ConfigEntry) -> None:
...
+    tracked_devices: set[str]
+    """Set of tracked device that will be used by the integration."""
+
+    name_prefix: str
+    """Name prefix to use for the entities."""
+
+    coordinator: ZaptecUpdateCoordinator
+    """The coordinator for the Zaptec account."""
+
+    streams: list[tuple[asyncio.Task, Installation]] = []
```
joostlek: Never initialize lists on class level, these are shared between all instances of this class

custom_components/zaptec/__init__.py
```diff
@@ -306,29 +409,50 @@ async def _first_time_setup(self) -> None:
...
             # These objects will be updated by the coordinator
-            self.updateable_objects = want
+            tracked_devices = keep
 
-        # Setup the stream subscription
-        self.create_streams()
+        return tracked_devices
+
+
+class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
```
joostlek, line 438: Move to coordinator.py

custom_components/zaptec/__init__.py
joostlek: I would argue that an entity description should not care into what kind of class it's thrown into

sveinse: If you look into any of sensor.py, button.py, number.py, you'll see that the list of entity descriptions includes `cls`. So adding `ZaptecEntityDescription` was just to provide a type definition that provides `cls`. Its actually never used anywhere for other than typing.

...I see now the docstring makes no sense. Maybe I'll reconsider the whole class.

sveinse: Yes, that's the plan. But in a later PR. I want to preserve the diff in this one. Otherwise the change will list as all new which is harder to review.

custom_components/zaptec/__init__.py
sveinse: It will be. The aim of this PR is to set the framework. The new implementation of all the necessary coordinators will be done in a later PR. This is to do this change piece-wise (which I assume is the practice in HA too).

custom_components/zaptec/__init__.py
sveinse: Yes. That's a copy-paste-typo. Thanks.

custom_components/zaptec/__init__.py
sveinse: Yes, that's the plan in later PR.

custom_components/zaptec/__init__.py
sveinse: Fixed.

sveinse:
@steinmn If you have the opportunity, it would be great if you could take a look at this. I have (finally) a working fix for the proper poll intervals, but I cannot add the PR before this has been merged.

custom_components/zaptec/__init__.py
```diff
@@ -112,31 +136,50 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
+        options=ZaptecUpdateOptions(
+            name="zaptec",
+            update_interval=scan_interval,
+            tracked_devices=manager.tracked_devices,
+            poll_args={"state": True, "info": True, "firmware": True},
+        ),
+    )
+    await manager.coordinator.async_config_entry_first_refresh()
+
+    # Install the local data for later retrieval
```
steinmn: "Store the manager in local runtime data for later retrieval"? This comment was a bit confusing to me

custom_components/zaptec/__init__.py
```diff
@@ -414,38 +561,21 @@ def _post_init(self) -> None:
...
-        _LOGGER.debug(
-            "    Added %s from %s",
-            self.entity_id,
-            self.zaptec_obj.qual_id,
-        )
-
     @callback
     def _handle_coordinator_update(self) -> None:
+        """Update the entity from Zaptec data."""
+        available = self._attr_available
```
steinmn:
```suggestion
        prev_available = self._attr_available
```

custom_components/zaptec/__init__.py
```diff
@@ -414,38 +561,21 @@ def _post_init(self) -> None:
...
+        update_from_zaptec = getattr(self, "_update_from_zaptec", lambda: None)
         try:
-            self._update_from_zaptec()
-        except Exception as exc:
-            raise HomeAssistantError(f"Error updating entity {self.key}") from exc
+            update_from_zaptec()
+            self._log_value(self._log_attribute)
+        except KeyUnavailableError as exc:
+            self._attr_available = False
+            self._log_unavailable(exc, available)
```
steinmn:
```suggestion
            self._log_unavailable(exc, prev_available)
```

custom_components/zaptec/__init__.py
```diff
@@ -414,38 +561,21 @@ def _post_init(self) -> None:
...
-        Called when the coordinator has new data. Implement this in the
-        inheriting class to update the entity state.
-        """
-
-    @callback
-    def _get_zaptec_value(self, *, default=MISSING, key=None):
+    def _get_zaptec_value(self, *, default=MISSING, key=None) -> Any:
         """Retrieve a value from the Zaptec object.
 
         Helper to retrieve the value from the Zaptec object. This is to
```
steinmn, line 590: Should we include that this function (and _update_from_zaptec) raises KeyUnavailableError in the docstring? Something along the lines of
```
Raises:
            KeyUnavailableError: If key doesn't exist or obj doesn't have `.get()`, which indicates that obj isn't a Mapping-like object
```
(According to https://github.com/home-assistant/architecture/discussions/878#discussioncomment-13662545, it should use google style for the docstring)

custom_components/zaptec/diagnostics.py
```diff
@@ -272,39 +296,39 @@ def addmap(k, v):
...
         )
     except Exception as err:
         add_failure(out, err)
 
     #
-    #  ENTITY MAP
+    #  ENTITIES
     #
     try:
+        # Clean up unused device entries with no entities
```
steinmn: Comment doesn't match code (copypaste from async_setup_entry?)

Issue #204, opened 2025-07-20
Lock type should not be used for permanent cable lock


sveinse:
According to @joostlek (from HA), the Lock entity type should only be used for physical locks that's used to enter the house. Not for locking the cable, as we do in this integration. The entity type needs to change to a switch.

PR #207, opened 2025-07-21
Change permanent_cable_lock from lock to switch

steinmn:
Fixes https://github.com/custom-components/zaptec/issues/204


sveinse:
Do we want to do this without any time of deprecation is does change the entity id from `lock.*_permanent_cable_lock` to `switch.*_permanent_cable_lock`.

Don't get me wrong, I'm not opposed to the idea, but do we want to break this? -- That being said, v0.8 is littered with behavioral changes, so now is maybe the time to do it? What do you think?

steinmn:
Well, #204 is listed both as a bug and as a part of the 0.8-milestone, so seemed like a good idea to just get it out of the way.

I have my charger in a locked garage, so I locked my cable once and haven't touched it since. Maybe other people use it more often, but it doesn't strike me as something you heavily integrate into a custom automation. And as you say, we have already removed a device and other significant changes for 0.8, so might as well keep going in my opinion.

That being said, I should probably add a point on it in the readme and the changelog.

sveinse:
@c0mplex1 you're not too late - its not out the door yet. If you only use the HA GUI to lock/unlock, you'll get the same functionality. It's the underlying name that changes since we changed from "lock" type to "switch" type. Functionally they are the same.

Issue #199, opened 2025-07-16
Reduce the number of calls to poll for firmware update


sveinse:
The `Charger.state()` function which polls for data updates are doing the following to poll for firmware updates:

https://github.com/custom-components/zaptec/blob/dce557e64e19ef3bf7adcab054075ac6e954e2f0/custom_components/zaptec/api.py#L622-L629

This is inefficient, as it will call this once per charger. The API end-point is for the installation, so the call only needs to happen once per installation for all the chargers it contain.

PR #208, opened 2025-07-21
Implemented differential poll intervals which significantly reduces polling

### LINKED ISSUE #205: Remove user settable scan/poll interval
HA integrations should not give the user the ability to set the scan/poll interval. We need to make the polling robust enough to work for all use cases of the integration without creating too many requests.

We need to remove this setting from our integration.

sveinse:
This PR implements the full system for differential poll intervals which significantly reduces the amount of polling. This fixes #205, #202, #199, #189 and is implementing as discussed in #188. Aims to address #165 

* Poll intervals adhere to the API guidelines
* Implemented multiple coordinators to handle the various poll intervals
  * Top which polls for firmware updates, once every day
  * Info which polls for device info updates, once every hour
  * One per device for handling run-time state data update. Every minute during charging, 10 minute otherwise
* Fixed trigger poll sequencer to use the coordinator and make sure only one run per device
* API: Added `Charger.is_charging()` method
* Removed user configurable scan interval

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -13,7 +13,6 @@ from typing import Any
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
     CONF_PASSWORD,
-    CONF_SCAN_INTERVAL,
     CONF_USERNAME,
     Platform,
 )
@@ -22,7 +21,6 @@ from homeassistant.exceptions import (
     ConfigEntryAuthFailed,
     ConfigEntryError,
     ConfigEntryNotReady,
-    HomeAssistantError,
 )
 from homeassistant.helpers import device_registry as dr, entity_registry as er
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
@@ -49,13 +47,16 @@ from .const import (
     CONF_CHARGERS,
     CONF_MANUAL_SELECT,
     CONF_PREFIX,
-    DEFAULT_SCAN_INTERVAL,
     DOMAIN,
     MANUFACTURER,
     MISSING,
     REQUEST_REFRESH_DELAY,
+    ZAPTEC_POLL_BUILD_INTERVAL,
     ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS,
+    ZAPTEC_POLL_CHARGING_INTERVAL,
+    ZAPTEC_POLL_INFO_INTERVAL,
     ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS,
+    ZAPTEC_POLL_IDLE_INTERVAL,
 )
 from .services import async_setup_services, async_unload_services
 
@@ -91,8 +92,10 @@ class ZaptecUpdateOptions:
 
     name: str
     update_interval: int
+    charging_update_interval: int | None
     tracked_devices: set[str]
     poll_args: dict[str, bool]
+    zaptec_object: ZaptecBase | None
 
 
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
@@ -113,14 +116,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     if prefix:
         prefix = prefix + " "
 
-    scan_interval = entry.data.get(CONF_SCAN_INTERVAL, DEFAULT_SCAN_INTERVAL)
-
     # Create the Zaptec object
     zaptec = Zaptec(
         entry.data[CONF_USERNAME],
         entry.data[CONF_PASSWORD],
         client=async_get_clientsession(hass),
-        max_time=scan_interval,
+        max_time=ZAPTEC_POLL_CHARGING_INTERVAL,  # The shortest of the intervals
     )
 
     # Login to the Zaptec account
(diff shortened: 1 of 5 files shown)


custom_components/zaptec/__init__.py
```diff
@@ -49,13 +47,16 @@
...
     DOMAIN,
     MANUFACTURER,
     MISSING,
     REQUEST_REFRESH_DELAY,
+    ZAPTEC_POLL_BUILD_INTERVAL,
     ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS,
+    ZAPTEC_POLL_CHARGING_INTERVAL,
+    ZAPTEC_POLL_INFO_INTERVAL,
     ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS,
+    ZAPTEC_POLL_IDLE_INTERVAL,
```
steinmn: Wouldn't it be nicer to keep interval in front of the different states, so that the intervals stick together when alfabetized? 
```suggestion
    ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS,
    ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS,
    ZAPTEC_POLL_INTERVAL_BUILD,
    ZAPTEC_POLL_INTERVAL_CHARGING,
    ZAPTEC_POLL_INTERVAL_INFO,
    ZAPTEC_POLL_INTERVAL_IDLE,
```

custom_components/zaptec/__init__.py
```diff
@@ -487,14 +591,15 @@ async def _trigger_poll(self, obj: ZaptecBase) -> None:
         HA initiated update.
         """
 
-        what = {obj.id}
         if isinstance(obj, Installation):
             delays = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
             kw = {"state": True, "info": True}
         else:
             delays = ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS
             kw = {"state": True}
```
steinmn: kw is no longer needed since it's now handled by self.options.poll_args

custom_components/zaptec/__init__.py
```diff
@@ -507,22 +612,35 @@ async def _trigger_poll(self, obj: ZaptecBase) -> None:
                 delta,
                 kw,
```
steinmn:
```suggestion
                self.options.poll_args,
```

custom_components/zaptec/__init__.py
```diff
@@ -465,11 +539,41 @@ def __init__(
...
+            options.zaptec_object, Charger
+        ):
+            self.async_add_listener(self.set_update_interval)
+
+    def set_update_interval(self) -> None:
+        """Set the update interval for the coordinator.
+
+        This function is called on data updates from the coordinator.
+        """
+        zaptec_obj = self.options.zaptec_object
```
steinmn: since `zaptec_obj.is_charging()` only exists if zaptec_obj is a Charger, I think we can make it explicit:
```suggestion
        zaptec_obj: Charger = self.options.zaptec_object
```

custom_components/zaptec/__init__.py
```diff
@@ -151,19 +152,66 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
+    )
+    # Dummy listener to ensure the coordinator runs
+    manager.info_coordinator.async_add_listener(lambda: None)
+
+    # Setup the device coordinators for each tracked device
+    for deviceid in tracked_devices:
+        zaptec_obj = zaptec[deviceid]
+
+        if isinstance(zaptec_obj, Installation):
+            poll_args = {"state": True, "info": True, "firmware": False}
```
steinmn: Does polling installation state do anything?

steinmn: And since installation info is already handled by info_coordinator, do we even want a timed polling of this here, or do we just want the ones triggered by HA-actions?

custom_components/zaptec/__init__.py
```diff
@@ -151,19 +152,66 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
+            poll_args = {"state": True, "info": False, "firmware": False}
+
+        manager.device_coordinators[deviceid] = ZaptecUpdateCoordinator(
+            hass,
+            entry=entry,
+            manager=manager,
+            options=ZaptecUpdateOptions(
+                name=deviceid,
+                update_interval=ZAPTEC_POLL_IDLE_INTERVAL,
+                charging_update_interval=ZAPTEC_POLL_CHARGING_INTERVAL,
```
steinmn: From here it looks like installation also changes polling frequency during charging, and it's not until you check in the ZaptecUpdateCoordinator constructor it becomes apparent that that is not the case.

custom_components/zaptec/__init__.py
sveinse: Very good idea. Fixed.

custom_components/zaptec/__init__.py
sveinse: Yes, that `kw` does nothing. Removed.

custom_components/zaptec/__init__.py
sveinse: No. kw, removed altogether. So fixed.

custom_components/zaptec/__init__.py
sveinse: Added. I also made the code even more defensive and raise an error if `zaptec_object` is anything else than a `Charger`. Its good to be explicit.

custom_components/zaptec/__init__.py
sveinse: Technically inthe info_coordinator and the device_coordinator overlaps on info polling for installations. My thinking is that the device coordinator is a replacement of not having state, and that we poll it every 10 minutes. The info coordinator is run every 60 minutes, so it doesn't overload the server having them both.

The great thing about having the device coordinator for the installation is that it automatically provides the mechanism for sequenced polls if the "max current" number dial is changed. Which I've seen many times is needed.

What this doesn't have thou, is your comment that triggers of changes in the installation device, should also trigger its chargers. That we'll have to look at later.

custom_components/zaptec/__init__.py
sveinse: Yes, I changed the code to make it more explicit. Please take a look and see if it is easier to follow now. Thanks.

custom_components/zaptec/__init__.py
steinmn: > What this doesn't have thou, is your comment that triggers of changes in the installation device, should also trigger its chargers. That we'll have to look at later.

As long as this is covered before we fully close #202, I think we're good here.

sveinse:
@steinmn 

> LGTM

Thank you.

> The only question I have is what happens if a periodic poll fails at some point, especially the long period ones?

The flippant answer is: The same way all other poll requests from HA are handled :D The more serious response is that I believe the coordinator have a mechanism for handling that, but I'm not 100% sure. It's a valid question.

> I occasionally get some `failed after 9 retries: Cannot connect to host api.zaptec.com:443 ssl:default [Timeout while contacting DNS servers]`-errors, and while that's probably just caused by my dev-container setup and/or running it in parallel with my regular HA-instance on the same network, I'm still curious how other errors would affect the system. Do we have some recovery mechanism that retries more often, or are the affected devices/entities now unavailable for the next full poll interval?

This is interesting, because say that there is some mechanism in coordinators that handles connection issues, that would not affect the other coordinators. They will keep on trying and eventually time out. This is something we definitely need to test more when the beta is out.

Issue #179, opened 2025-05-25
Load balancing blueprint issue


Ventilgummi:
Hi, there's an issue with the blueprint for loadbalancing.
I've solved it by renaming my entities, but could be improved from the blueprint side.
See my closed issue here: https://github.com/svenakela/ha/issues/9

steinmn:
> but could be improved from the blueprint side.
> See my closed issue here: [svenakela/ha#9](https://github.com/svenakela/ha/issues/9)

@Ventilgummi I agree that this could be improved from the blueprint side, so I suggest you close this issue and reopen https://github.com/svenakela/ha/issues/9 instead (or create a new issue in [svenakela/ha](https://github.com/svenakela/ha/issues/new/choose)), since that is where it should be improved/fixed.

PR #215, opened 2025-07-22
Don't allow charging button to be pushed unless the command is valid

### LINKED ISSUE #209: Charging switch doesn't work when authentication is required
When using a system that requires authentication, the charger might be in "Waiting" mode. When the Charging switch is pressed it fails and is stuck in "on" mode forever. It doesn't go back when the charger is polled for data.

The errors from the log:

```
2025-07-22 11:54:43.787 WARNING (MainThread) [custom_components.zaptec.api] Resume charging is not allowed if charger is not paused
2025-07-22 11:54:43.787 DEBUG (MainThread) [custom_components.zaptec.api] operation_mode: Connected_Requesting, final_stop_active: 0
2025-07-22 11:54:43.787 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [131416884311552] Error during service call to switch.turn_on: Resuming charging failed
```

Turning it off sends a "stop_charging_final" command to the charger which succeeds. When that happens two entity updates are seen:

```
2025-07-22 12:05:02.437 DEBUG (MainThread) [custom_components.zaptec.api] >>>     Updating Charger[f40861].final_stop_active (FinalStopActive)  =  <str> 1  (was 0)
2025-07-22 12:05:07.488 DEBUG (MainThread) [custom_components.zaptec.api] >>>     Updating Charger[f40861].charger_operation_mode (ChargerOperationMode)  =  <str> Connected_Finished  (was Connected_Requesting)
```

PS! The entity "Charger mode" is still stuck in "Waiting" after this. This is possibly a separate issue, but I'm noting it here so it can be investigated and forked off to a separate issue if needed.

### LINKED ISSUE #212: Charger charging state is not indicated in HA
As suspected in #209:

My charger doesn't indicate charging at all when charging, nor does it decrease the poll rate while charging. This can be read from the debug logs:

```
2025-07-22 12:27:17.886 DEBUG (MainThread) [custom_components.zaptec.api] ---   Subscription: {'DeviceId': 'xxx', 'DeviceType': 4, 'ChargerId': 'xxxx', 'StateId': '710 (ChargerOperationMode)', 'Timestamp': '2025-07-22T12:27:17.412842Z', 'ValueAsString': '3'}
2025-07-22 12:27:17.886 DEBUG (MainThread) [custom_components.zaptec.api] >>>     Updating Charger[f40861].charger_operation_mode (ChargerOperationMode)  =  <str> Connected_Charging  (was Connected_Requesting)
```

I'm also observing the following on the next poll:

```
2025-07-22 12:27:21.432 DEBUG (MainThread) [custom_components.zaptec.api] Poll state from Charger[f40861] (Sol)
2025-07-22 12:27:21.476 DEBUG (MainThread) [custom_components.zaptec.api] >>>     Updating Charger[f40861].charge_mode (ChargeMode)  =  <str> 6  (was 9)
```

I suspect there are two things here:

1. We should probably use `ChargerOperationMode` as data source for our indication of charge state, since this is updated by the service bus
2. It seems the number to string translation in `ChargeMode` is not working right, and probably the cause for the observed issue.

### LINKED ISSUE #213: Change source of charge mode in HA
With respect to #209 and #212, the attribute used in HA to indicate charger mode must change. The current field `ChargerMode` is updated via info updates, not via state. Instead `ChargerOperationMode` is a state variable which we should use.

steinmn:
Addresses #209 by disabling the button if in an invalid state

Fixes #209, fixes #212, fixes #213

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -266,9 +266,14 @@ def remove_deprecated_entities(hass: HomeAssistant, entry: ZaptecConfigEntry) ->
     """Remove deprecated entites from the entity_registry"""
 
     entity_registry = er.async_get(hass)
-    zaptec_entity_list = [(entity_id, entity) for entity_id, entity in list(entity_registry.entities.items()) if entity.config_entry_id == entry.entry_id]
+    zaptec_entity_list = [(entity_id, entity) for entity_id, entity in list(
+        entity_registry.entities.items()) if entity.config_entry_id == entry.entry_id]
     for entity_id, entity in zaptec_entity_list:
         if entity.translation_key == 'operating_mode':
+            # The two entities this applies to were changed to the key
+            # charger_operation_mode (from state) instead of operating_mode (from info).
+            # In order to keep the same entity_id, we need to remove the old entries
+            # before the new ones are added.
             _LOGGER.warning("Removing deprecated entity: %s", entity_id)
             entity_registry.async_remove(entity_id)
 
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -103,6 +103,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     _LOGGER.debug("Setting up entry %s: %s", entry.entry_id, redacted_data)
 
+    # Remove deprecated entities where we need to reuse the entity_ids
+    remove_deprecated_entities(hass, entry)
+
     configured_chargers = None
     if entry.data.get(CONF_MANUAL_SELECT, False):
         configured_chargers = entry.data.get(CONF_CHARGERS)
@@ -257,6 +260,17 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
                 device_registry.async_remove_device(dev.id)
 
     return True
+
+
+def remove_deprecated_entities(hass: HomeAssistant, entry: ZaptecConfigEntry) -> None:
+    """Remove deprecated entites from the entity_registry"""
+
+    entity_registry = er.async_get(hass)
+    zaptec_entity_list = [(entity_id, entity) for entity_id, entity in list(entity_registry.entities.items()) if entity.config_entry_id == entry.entry_id]
+    for entity_id, entity in zaptec_entity_list:
+        if entity.translation_key == 'operating_mode':
+            _LOGGER.warning("Removing deprecated entity: %s", entity_id)
+            entity_registry.async_remove(entity_id)
 
 
 async def async_unload_entry(hass: HomeAssistant, entry: ZaptecConfigEntry) -> bool:
--- a/custom_components/zaptec/switch.py
+++ b/custom_components/zaptec/switch.py
@@ -42,6 +42,12 @@ class ZaptecChargeSwitch(ZaptecSwitch):
     zaptec_obj: Charger
     _log_attribute = "_attr_is_on"
 
+    @property
+    def available(self) -> bool:
+        """Return True if entity is available."""
+        command = "stop_charging_final" if self._attr_is_on else "resume_charging"
+        return super().available and self.zaptec_obj.is_command_valid(command)
+
     @callback
     def _update_from_zaptec(self) -> None:
         """Update the entity from Zaptec data."""
--- a/custom_components/zaptec/switch.py
+++ b/custom_components/zaptec/switch.py
@@ -53,7 +53,7 @@ class ZaptecChargeSwitch(ZaptecSwitch):
         """Update the entity from Zaptec data."""
         # Called from ZaptecBaseEntity._handle_coordinator_update()
         state = self._get_zaptec_value()
-        self._attr_is_on = state in ["Connected_Charging"]
+        self._attr_is_on = state in ["Connected_Charging", "Connected_Requesting"]
         self._attr_available = True
 
     async def async_turn_on(self, **kwargs):  # pylint: disable=unused-argument
--- a/custom_components/zaptec/switch.py
+++ b/custom_components/zaptec/switch.py
@@ -53,7 +53,7 @@ class ZaptecChargeSwitch(ZaptecSwitch):
         """Update the entity from Zaptec data."""
         # Called from ZaptecBaseEntity._handle_coordinator_update()
         state = self._get_zaptec_value()
-        self._attr_is_on = state in ["Connected_Charging", "Connected_Requesting"]
+        self._attr_is_on = state in ["Connected_Charging"]
         self._attr_available = True
 
     async def async_turn_on(self, **kwargs):  # pylint: disable=unused-argument
--- a/custom_components/zaptec/switch.py
+++ b/custom_components/zaptec/switch.py
@@ -134,8 +134,8 @@ INSTALLATION_ENTITIES: list[EntityDescription] = []
 
 CHARGER_ENTITIES: list[EntityDescription] = [
     ZapSwitchEntityDescription(
-        key="operating_mode",
-        translation_key="operating_mode",
+        key="charger_operation_mode",
+        translation_key="charger_operation_mode",
         device_class=SwitchDeviceClass.SWITCH,
         cls=ZaptecChargeSwitch,
     ),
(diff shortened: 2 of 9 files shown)


steinmn:
Added fix for #212 and #213. A minor issue is that this creates new entities with <oldname>_2 for the charging mode sensor and switch. Should  we try to handle this automatically somehow?

steinmn:
If you are talking about entity names, `*_charger_mode` is the snake_case resulting from the name of the sensor/operating_mode entry in en.json. Different languages get different entity_ids (ref  #179 and https://github.com/svenakela/ha/issues/9)

sveinse:
> If you are talking about entity names, `*_charger_mode` is the snake_case resulting from the name of the sensor/operating_mode entry in en.json. Different languages get different entity_ids (ref #179 and [svenakela/ha#9](https://github.com/svenakela/ha/issues/9))

Yes, with the latest fix in PR #216 the log now prints the Zaptec attribute name used in the entity, so that the mapping from Zaptec attribute to entity name is clearer.

> Added fix for #212 and #213. A minor issue is that this creates new entities with _2 for the charging mode sensor and switch. Should we try to handle this automatically somehow?

We could "find" the old entity in the end of `async_setup_entry()` in __main__.py and delete it. If we happen to do that _before_ the new entity is added, do they get the same name? Haven't tested it, and it is possibly somewhat hacky. I'm not sure when its ok to go about and delete entities and when its not.

sveinse:
There is another option:

Today we use the `EntityDescription.key` as both the entity creation key (which HA handles) and the Zaptec lookup key. What if we separate them? Make an optional field `zaptec_key` which those few entities that needs it can use. This would allow us to change the underlying Zaptec reference without affecting the HA entity show.

If we go this route, we need a dedicated issue and PR, as this is a new "feature".

steinmn:
Entities do get the same name/entity_id when the old entity is deleted before we add the new ones. I think there is already enough confusion with all the different ids and keys without adding yet another one, so I think I prefer the removal option.

steinmn:
Are we dealing with that problem as part of 0.8?

custom_components/zaptec/__init__.py
```diff
@@ -259,6 +262,17 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     return True
 
 
+def remove_deprecated_entities(hass: HomeAssistant, entry: ZaptecConfigEntry) -> None:
+    """Remove deprecated entites from the entity_registry"""
+
+    entity_registry = er.async_get(hass)
+    zaptec_entity_list = [(entity_id, entity) for entity_id, entity in list(entity_registry.entities.items()) if entity.config_entry_id == entry.entry_id]
+    for entity_id, entity in zaptec_entity_list:
+        if entity.translation_key == 'operating_mode':
```
sveinse, line 272: I think this deserves a comment that this entity will be readded again and that this is done to change the back-end reference to Zaptec.

Have you tested that your automations is kept intact when you do this?

steinmn: No, I haven't this automation to my dev-container yet. I can test it tomorrow once the sun is up. `{% set charger_state = states('sensor.*_charger_mode') %}` should still work though, the comparisons against 'Charging' and 'Waiting' are the ones that will break, since they should now compare against 'Connected_Charging' and 'Connected_Requesting' respectively, see earlier comments. Is this too much of a breaking change, or is it ok?

steinmn: (updated comment)

custom_components/zaptec/switch.py
```diff
@@ -42,6 +42,12 @@ class ZaptecChargeSwitch(ZaptecSwitch):
     zaptec_obj: Charger
     _log_attribute = "_attr_is_on"
 
+    @property
+    def available(self) -> bool:
+        """Return True if entity is available."""
+        command = "stop_charging_final" if self._attr_is_on else "resume_charging"
+        return super().available and self.zaptec_obj.is_command_valid(command)
```
sveinse, line 49: It might be interesting to know that I found out today during testing that for chargers with require authorization, then this is the flow to start it (again):

Connected_Finished --> Command "resume_charging" --> Connected_Requesting --> Command "authorize_charge" --> Connected_Charging.

Point is, the "resume_charging" command might not take it to charging mode. So users might see that the switch flips back.

I'm not convinced solving that is a topic for this PR.

steinmn: The simplest solution would be to have Connected_Requesting lead to an ON state for the switch. I think this might also be more correct than the current situation, at least it corresponds better with when resume/stop commands are allowed.

steinmn: In other words, changing line 56 to
```self._attr_is_on = state in ["Connected_Charging", "Connected_Requesting"]```

Comment (steinmn) [reply custom_components/zaptec/switch.py:49]: Added this as well as an update to the readme.

**Comment by steinmn** [comment]
The computation part of my automation works fine in a minimal test (once I changed `Waiting` to `Connected_Requesting`and `Charging` to `Connected_Charging` as expected). Haven't run the automation closed loop, both because I couldn't install the tibber-integration in the dev-container (dependency issue I'll leave for another day in favor of hiking in the nice weather), and because I rely on adjusting available_current on the Installation so I need #220 resolved for it to work properly.

I consider this ready for review/testing

File: custom_components/zaptec/switch.py (hunk 2 above)
Comment (sveinse) [reply custom_components/zaptec/switch.py:49]: Hmm. What is this switch supposed to indicate? Technically "Connected_Requesting" isn't charging your car.

What I meant was that using "resume_charging" might not in all use-cases result in the car charging. If we are to use this switch in those use-cases AND use the switch to turn on the charging, we need a sequence of commands to take it through the states. I'm very skeptical of adding this logic thou.

File: custom_components/zaptec/switch.py [hunk 3]
```diff
@@ -53,7 +53,7 @@ def _update_from_zaptec(self) -> None:
         """Update the entity from Zaptec data."""
         # Called from ZaptecBaseEntity._handle_coordinator_update()
         state = self._get_zaptec_value()
-        self._attr_is_on = state in ["Connected_Charging"]
+        self._attr_is_on = state in ["Connected_Charging", "Connected_Requesting"]
```
sveinse: I think this will be confusing for the user. HA indicates charging, while the car is not charging.

Without it, the switch will show not charging if the user tries to start it (without authentication), but its reflecting the actual state of charging. It's not ideal, but I think this is better. I recommend reverting this one.

steinmn: Done

I think there are too many different ways charging can be blocked that is unaffected by calling the `resume_charging`-command to try to make this button make any more sense. The important part is that users will no longer be allowed to click the switch if the state is `Connected_Requesting`,  which should lead them to check one of the other potential blockers.

custom_components/zaptec/switch.py
steinmn: > What is this switch supposed to indicate?

Yeah, this button in general doesn't really make much sense to me, so I think I tried it maybe once or twice after the first time I installed the integration, and then just started using `available_current` instead.
> I'm very skeptical of adding this logic thou.

Yep, looks like a recipe for tons of bugs, pain and regret. Far too many options to keep track of (`available_current`, `charger_max_current`, various authentication approaches and internal scheduling in the car are just a couple off the top of my head)

custom_components/zaptec/switch.py
sveinse: Thanks.

Personally I never use this switch. My automation looks at the charger state and issue the specific commands (via the buttons). During a ongoing session, charging is controlled via the current settings and paused by setting it to zero.

I think the intended audience of this switch are the "standard" users with lesser experience. So perhaps we should write in the docs that "The charge switch is a basic function that works for the simple use cases. In more advanced setups, such as with authentication, this button might not start the charging immediately due to the extra steps needed for authentication." Something like that?

steinmn: Kind of already says that in https://github.com/custom-components/zaptec?tab=readme-ov-file#start--stop-charging, doesn't it? Maybe expand the caveats to include current limit in addition to authentication?

sveinse (approved):
LGTM.

If you're all set, I will proceed with the merge.

PR #218, opened 2025-07-22
Add setting to enable if entity is to be checked on startup

### LINKED ISSUE #217: Entities won't get added if they miss from Zaptec
Testing a restart of HA revealed the following exception:

```
2025-07-22 15:21:00.579 ERROR (MainThread) [custom_components.zaptec] Failed to add entity ZaptecSensor keys total_charge_power_session, skipping entity
  File "/workspaces/ha-core/config/custom_components/zaptec/__init__.py", line 367, in create_entities_from_descriptions
    updater()
    ~~~~~~~^^
  File "/workspaces/ha-core/config/custom_components/zaptec/sensor.py", line 35, in _update_from_zaptec
    self._attr_native_value = self._get_zaptec_value()
                              ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/workspaces/ha-core/config/custom_components/zaptec/__init__.py", line 731, in _get_zaptec_value
    raise KeyUnavailableError(
        f"Failed to retrieve {key!r} from {self.zaptec_obj.qual_id}. Key {k!r} doesn't exist"
    )
custom_components.zaptec.KeyUnavailableError: Failed to retrieve 'total_charge_power_session' from Charger[f40861]. Key 'total_charge_power_session' doesn't exist
```

This happens because this attribute is not always available at all times from Zaptec. PR #203 added that the attribute was checked during startup and refrain from adding the entity if it fails, which is this exception. The reason for adding this was that it's the recommended behavior by HA. BUT only if there is a mechanism for adding it run-time when it is available again, which we don't have.

I think we should revert or comment out this code:

https://github.com/custom-components/zaptec/blob/master/custom_components/zaptec/__init__.py#L361-L374

sveinse:
It turns out that some Zaptec attributes are not always present from Zaptec. An example is "total_charge_power_session". The recommendation from HA is to not add entities that aren't available, but we have to circumvent that in this case. Instead a setting have been added so it can be later turned on if wanted.

Fixes #217

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -44,6 +44,7 @@ from .const import (
     CONF_MANUAL_SELECT,
     CONF_PREFIX,
     DOMAIN,
+    KEYS_TO_SKIP_ENTITY_AVAILABILITY_CHECK,
     MANUFACTURER,
     MISSING,
     REQUEST_REFRESH_DELAY,
@@ -68,15 +69,6 @@ PLATFORMS = [
     Platform.SWITCH,
     Platform.UPDATE,
 ]
-
-CHECK_ENTITY_AVAILABLE_ON_ADD = False
-"""Enable a check to ensure that the entity is available when added.
-
-If not available, the entity will not be added to the list of entities.
-For Zaptec, this is not recommended, as some attributes are not always
-available. E.g. "TotalChargePowerSession" is only present under certain
-conditions.
-"""
 
 class KeyUnavailableError(Exception):
     """Exception raised when a key is not available in the Zaptec object."""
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/__init__.py
```diff
@@ -69,6 +69,14 @@
     Platform.UPDATE,
 ]
 
+CHECK_ENTITY_AVAILABLE_ON_ADD = False
+"""Enable a check to ensure that the entity is available when added.
+
+If not available, the entity will not be added to the list of entities.
+For Zaptec, this is not recommended, as some attributes are not always
+available. E.g. "TotalChargePowerSession" is only present under certain
+conditions.
```
steinmn: Does "some" mean 1-2 or significantly more? In other words: do we want to disable this for all entities, or is there a limited number this applies to?

sveinse: _I_ see this only for "TotalChargePowerSession". But there is no spec or description from Zaptec so we can't know. My thinking is that if this can happen to one, it can happen to many. Should we rather be empirical and do something special when we encounter them instead?

Heh, I'm thinking about my kid using "everyone in class is..." to argue something is needed :D

steinmn: Even if we don't add them, they are not removed from the entity registry, right?

So if we have a list of the entities we want to skip the check for in the first 8.0-beta, we can include instructions for beta testers to look out for these entities "no longer being provided by the zaptec integration." and the entities will return safely after an updated beta with an extended list?

sveinse: No, they wont get deleted. But they won't be added either. We're lacking a mechanism to monitor when a zaptec attribute is being added, so we cant create the entity run-time atm.

"TotalChargePowerSession" is something I frequently use. I need a fix for this one at least.

steinmn: So could something like `if description.key not in KEYS_TO_SKIP_ENTITY_AVAILABILITY_CHECK_FOR` where `KEYS_TO_SKIP_ENTITY_AVAILABILITY_CHECK_FOR = ["total_charge_power_session"]` work?

sveinse: Yes, that can work. That list should be placed in const.py with clear documentation on what it is. It becomes a middle ground between all or nothing.

sveinse:
@steinmn I updated according to your comments, and I think this looks better now.

There is one remaining item I'd like to add. There is a guideline in HA that errors and warnings should not be produced when it's expected, so I need to change it to not log as warning or error when it an entity we know can go away.

PR #225, opened 2025-07-23
Fix missing attribute types

### LINKED ISSUE #222: Attributes missing proper datatype
These entities are getting strange types from our API:

```
sensor.sol_humidity  =  '36.892' <str>   from Charger[f40861].humidity
sensor.sol_temperature_internal  =  '33.019' <str>   from Charger[f40861].temperature_internal5
sensor.sol_allocated_charge_current  =  '12' <str>   from Charger[f40861].charge_current_set
```

I actually think HA is ok with that, but there is a type conversion which isn't working right here.

steinmn:
These are the attributes directly used in entities. There are still some more internal attributes that we could consider adding typing for, but I think I'll leave it like this for now

Fixes #222


sveinse (approved):
LGTM

PR #226, opened 2025-07-23
Trig all sub chargers on installation trigger

### LINKED ISSUE #220: Trigger polling of child chargers for installation commands
(Some) installation commands should also trigger polls of associated chargers. For instance, updating available_current for an installation will most likely change the charge power of the chargers in it.

(leftover from #202)

sveinse:
This PR ensures that chargers also gets triggered when a command is issued to installation. The typical use case is if the available current in the charger has changed, this might affect the current in the individual chargers.

Fixes #220

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -593,8 +593,16 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
         HA initiated update.
         """
 
+        coordinators: list[ZaptecUpdateCoordinator] = [self]
         if isinstance(obj, Installation):
             delays = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
+            # If the installation has chargers, we also trigger the
+            # coordinators for the chargers that are tracked.
+            coordinators += [
+                self.manager.device_coordinators[charger.id]
+                for charger in obj.chargers
+                if charger.id in self.manager.tracked_devices
+            ]
         else:
             delays = ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS
 
@@ -611,7 +619,8 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
                 obj.qual_id,
                 delta,
             )
-            await self.async_refresh()
+            for coord in coordinators:
+                await coord.async_refresh()
 
     async def trigger_poll(self, obj: ZaptecBase) -> None:
         """Trigger a poll update sequence."""
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -1,5 +1,6 @@
 """Zaptec component."""
 
+
 from __future__ import annotations
 
 import asyncio
@@ -68,7 +69,6 @@ PLATFORMS = [
     Platform.SWITCH,
     Platform.UPDATE,
 ]
-
 
 
 class KeyUnavailableError(Exception):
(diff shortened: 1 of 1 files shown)


steinmn (commented):
One issue here is that the child chargers will be polled with the installation trigger delays (ie without the last trigger after 15 seconds)

Wouldn't it be better to have trigger_poll (no leading underscore) call the trigger_poll-functions of the respective coordinators after setting up the background task? Then we also avoid potentially having both the installation-coordinator and the charger-coordinator running a polling task for the same charger if the user has changed something on both the charger and the installation within the 15second timeperiod.

sveinse:
Yes, I saw the same thing with the 15 second delay missing. You're right. It's better to trigger them the other way. I'll fix it. Thanks.

sveinse:
@steinmn This is ready for review.

This PR leaves service calls broken. #229. I intend to fix this in a separate PR.

custom_components/zaptec/__init__.py
```diff
@@ -608,19 +617,28 @@ async def _trigger_poll(self, obj: ZaptecBase) -> None:
...
-                obj.qual_id,
+                zaptec_obj.qual_id,
                 delta,
             )
             await self.async_refresh()
 
-    async def trigger_poll(self, obj: ZaptecBase) -> None:
+            # Trigger the poll for the children coordinators
+            for coord in children_coordinators:
+                await coord.trigger_poll()
```
steinmn: This is inside the for loop, so charger(s) will now trigger 4 times: 4, 9, 14 and 22 seconds after the triggering event. This is because the first round of polling is interrupted by a new trigger from the installation poller

custom_components/zaptec/__init__.py
```diff
@@ -636,8 +654,8 @@ def cleanup_task(_task: asyncio.Task):
 
         self._trigger_task = self.config_entry.async_create_background_task(
             self.hass,
-            self._trigger_poll(obj),
-            f"Zaptec Poll Update for {obj.qual_id}",
+            self._trigger_poll(zaptec_obj),
+            f"Zaptec Poll Update for {zaptec_obj.qual_id}",
         )
         self._trigger_task.add_done_callback(cleanup_task)
```
steinmn, line 661: After this line was where I thought it would make sense to add `for coord in children_coordinators`.

custom_components/zaptec/__init__.py
```diff
@@ -586,19 +585,29 @@ async def _async_update_data(self) -> None:
...
+        children_coordinators: list[ZaptecUpdateCoordinator] = []
+        if isinstance(zaptec_obj, Installation):
             delays = ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS
+            # If the installation has chargers, we also trigger the
+            # coordinators for the chargers that are tracked.
+            children_coordinators = [
+                self.manager.device_coordinators[charger.id]
+                for charger in zaptec_obj.chargers
+                if charger.id in self.manager.tracked_devices
+            ]
```
steinmn, line 604: Do we need to do this here to be prepared for changes, or could we do this at the end of create_entities_from_descriptions?

steinmn: Sorry, async_setup_entry was what I should have said

sveinse: Yeah, so we set it up as a field in ZaptecUpdateOptions? Not a bad idea. I'll sit on this. One immediate thing I can think of is if we add/remove chargers on the fly, we have another field to keep updated. But we don't so maybe this is a non-issue atm.

custom_components/zaptec/__init__.py
sveinse: Yes, I didn't catch this. I'll fix this as soon as I'm done with the PR I'm working on.

sveinse: Fixed

custom_components/zaptec/__init__.py
sveinse: No. I'd like to have it called after the installation trigger has run once. You can cee my fix in the latest patch.

custom_components/zaptec/__init__.py
sveinse: I think I'll leave it like this for now.

sveinse:
@steinmn This is ready as well. Thanks for reviewing.

PR #228, opened 2025-07-23
Update devcontainer to latest debian version

steinmn:
Main reason for updating is to get SQLite 3.40.1 (old version had 3.34.1). This was preventing the recorder setup, which in turn prevented the setup of energy, history and logbook and the installation of any integration relying on any of them (for instance the tibber-integration).


sveinse:
It looks good, but I do not have any opinions on this. If you think this is useful, I will approve and merge.

steinmn:
Yes, it is a simple fix for the tibber-issue I mentioned I had on a different PR, and it keeps the devcontainer up to date and compatible with newer ha versions.

Issue #219, opened 2025-07-22
Entity are now from Zaptec sources that updated slowly


sveinse:
With the introduction of v0.8, the poll rates for getting charger "info" is being significantly reduced. Instead one should use "state".

These sensors are not updated with the "quick" state poll which is running at either 10 minute or 1 minute (depending on charger state).

#### From info (refresh rate: 60 minutes)
* switch.sol_charging from {charger}.operating_mode
* binary_sensor.sol_online from {charger}.is_online
* binary_sensor.sol_authorization_required from {charger}.is_authorization_required
* sensor.sol_charger_mode from {charger}.operating_mode
* sensor.sol_energy_meter from {charger}.signed_meter_value_kwh

#### From build (refresh rate: only on startup)
* binary_sensor.sol_charger from {charger}.active
* sensor.sol_device_type from {charger}.device_type

steinmn:
Ok

IsOnline has stateid=-2, so that is already covered. 
DeviceType does not strike me as something that will ever change.
Is Active something that changes?

steinmn:
Yes, the signed meter value sounds like it should be its' own issue.

PR #230, opened 2025-07-23
Use state value for authorization required for more frequent updates

steinmn:
Use `authentication_required` instead of `is_authorization_required` since that is part of the state update.

Fixes (part of) #219

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -283,6 +283,14 @@ def remove_deprecated_entities(hass: HomeAssistant, entry: ZaptecConfigEntry) ->
             # before the new ones are added.
             _LOGGER.warning("Removing deprecated entity: %s", entity_id)
             entity_registry.async_remove(entity_id)
+        elif entity.unique_id.endswith("_is_authorization_required"):
+            # There is an entity using authorization_required as a translation key in
+            # both the installation and the charger device. We only want to replace the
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/__init__.py
```diff
@@ -283,6 +283,14 @@ def remove_deprecated_entities(hass: HomeAssistant, entry: ZaptecConfigEntry) ->
...
             _LOGGER.warning("Removing deprecated entity: %s", entity_id)
             entity_registry.async_remove(entity_id)
+        elif entity.unique_id.endswith("_is_authorization_required"):
+            # There is an entity using authorization_required as a translation key in
+            # both the installation and the charger device. We only want to replace the
+            # entity associated with the charger, so we use the end of the unique_id
+            # to find the correct entity that will be readded later (the installation
+            # entity ends with '_is_required_authentication').
+            _LOGGER.warning("Removing deprecated entity: %s", entity_id)
+            entity_registry.async_remove(entity_id)
```
sveinse: Looks good. I wonder thou: Should we add to the comment of this one and the "operating_mode" that this added for the upgrade from "v0.7 -> v0.8"? This is useful information to have when this is reviewed in the future and the memory of why this code exists is forgotten.

steinmn: While git-blame does help with this, that is a VERY good point, will add.

sveinse (approved):
LGTM. Thank you.

PR #235, opened 2025-07-23
Change data source for charger energy meter

### LINKED ISSUE #219: Entity are now from Zaptec sources that updated slowly

### LINKED ISSUE #233: Change source of entity "energy_meter" to use state variable
With respect to #219, see comment https://github.com/custom-components/zaptec/issues/219#issuecomment-3109798868

The integration use "signed_meter_value_kwh" as source which is not a state variable. Instead there is "signed_meter_value", but this is encoded using Open Charge Metring Format (OCMF) which is described here:  https://github.com/SAFE-eV/OCMF-Open-Charge-Metering-Format/blob/master/OCMF-en.md

Looks like this

```json
{
  "FV": "1.0",
  "GI": "ZAPTEC GO",
  "GS": "ZAPxxxxxx",
  "GV": "2.4.2.4",
  "PG": "F1",
  "RD": [
    {
      "TM": "2025-07-22T20:00:00,000+00:00 R",
      "RV": 6211.03,
      "RI": "1-0:1.8.0",
      "RU": "kWh",
      "RT": "AC",
      "ST": "G"
    }
  ]
}
```

What's useful about this is that this is pushed regularly via the service bus, so it helps keeping the data in HA up to date.

sveinse:
* Create new ZaptecEnergySensor class
* Implement a simple OCMF decoder to extract the latest energy meter reading

Fix #233, Fix #219

### DIFF
--- a/custom_components/zaptec/misc.py
+++ b/custom_components/zaptec/misc.py
@@ -91,13 +91,13 @@ def mc_nbfx_decoder(msg: bytes) -> None:
             raise AttributeError(f"Unknown record type {hex(record_type)}")
 
 
-def get_ocmf_max_reader_value(data: dict) -> int:
+def get_ocmf_max_reader_value(data: dict) -> float:
     """Return the maximum reader value from OCMF data."""
 
     if not isinstance(data, dict):
-        return 0
+        return 0.
     return max(
-        int(reading.get("RV", 0))
+        float(reading.get("RV", 0.))
         for reading in data.get("RD",[])
     )
 
--- a/custom_components/zaptec/misc.py
+++ b/custom_components/zaptec/misc.py
@@ -1,7 +1,6 @@
 """Misc helper stuff."""
 from __future__ import annotations
 
-from datetime import datetime, timedelta, timezone
 import re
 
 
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/misc.py
```diff
@@ -91,6 +92,48 @@ def frame_decoder():
...
+                int(m[7]) * 1000,  # Milliseconds
+                tzinfo=timezone(timedelta(hours=int(m[8]), minutes=int(m[9]))),  # Time zone
+            )
+            readings.append((ts, value))
+
+    # Sort readings by timestamp
+    readings.sort(key=lambda x: x[0])
+    if readings:
+        return readings[-1][1]  # Return the latest value
+    return 0
```
steinmn: Aren't we overcomplicating this? This is an increasing value that will never decrease, isn't it? So we just need the maximum value, right?

custom_components/zaptec/sensor.py
```diff
@@ -220,13 +251,13 @@ class ZapSensorEntityDescription(SensorEntityDescription):
         cls=ZaptecSensor,
     ),
     ZapSensorEntityDescription(
-        key="signed_meter_value_kwh",
+        key="signed_meter_value_kwh",  # This key is unused, but kept for <0.7 compatibility
```
steinmn:
```suggestion
        key="signed_meter_value_kwh",  # This key is no longer used to get Zaptec values, but is linked to the entity unique_id, so it is kept for <0.7 compatibility
```

steinmn (commented):
This should work as written, but I think it can be simplified by ignoring time and just using the max value.

custom_components/zaptec/misc.py
sveinse: Yes, I am. Fixed now.

custom_components/zaptec/sensor.py
sveinse: Added. Thanks.

custom_components/zaptec/misc.py
```diff
@@ -91,6 +91,17 @@ def frame_decoder():
             raise AttributeError(f"Unknown record type {hex(record_type)}")
 
 
+def get_ocmf_max_reader_value(data: dict) -> int:
```
steinmn:
```suggestion
def get_ocmf_max_reader_value(data: dict) -> float:
```

custom_components/zaptec/misc.py
```diff
@@ -91,6 +91,17 @@ def frame_decoder():
...
+    """Return the maximum reader value from OCMF data."""
+
+    if not isinstance(data, dict):
+        return 0
+    return max(
+        int(reading.get("RV", 0))
```
steinmn:
```suggestion
        float(reading.get("RV", 0))
```

steinmn (commented):
The energy meter values should be floats, otherwise LGTM

steinmn (approved):

PR #237, opened 2025-07-24
Fixup service/actions and deprecate old

### LINKED ISSUE #229: Commands issued via the service call is broken
Due to the changes in the way polling is done, the way service call commands are trigging updates is broken.

```python
    async def service_handle_stop_charging(service_call: ServiceCall) -> None:
        _LOGGER.debug("Called stop charging")
        for coordinator, obj in iter_objects(service_call, Charger):
            _LOGGER.debug("  >> to %s", obj.id)
            try:
                await obj.command("stop_charging_final")
            except Exception as exc:
                raise HomeAssistantError(
                    f"Command 'stop_charging_final' failed: {exc}"
                ) from exc
            await coordinator.trigger_poll(obj)    # <-- This is not working
```

PS! A timely question: Should calling the commands via the service be deprecated? I think we should use service only in the cases where there are special cases and conditions that the entities cannot provide. If we want the user to send arbitrary commands to the charger, then I think a generic "service_command" would be better.

sveinse:
* Fixup sevice/action calls which didn't work after v0.8 change
* Mark commands which have button entities as deprecated

Fixes #229

### DIFF
--- a/custom_components/zaptec/services.py
+++ b/custom_components/zaptec/services.py
@@ -202,7 +202,7 @@ async def async_setup_services(hass: HomeAssistant, manager: ZaptecManager) -> N
         _LOGGER.debug("Called stop charging")
         _LOGGER.warning(
             "The 'stop_charging' action is deprecated and will be removed in a future release. "
-            "Use the 'stop_charging' botton entity instead"
+            "Use the 'Stop charging' button entity instead"
         )
         for coordinator, obj in iter_objects(service_call, Charger):
             _LOGGER.debug("  >> to %s", obj.id)
@@ -218,7 +218,7 @@ async def async_setup_services(hass: HomeAssistant, manager: ZaptecManager) -> N
         _LOGGER.debug("Called resume charging")
         _LOGGER.warning(
             "The 'resume_charging' action is deprecated and will be removed in a future release. "
-            "Use the 'Resume Charging' button entity instead"
+            "Use the 'Resume charging' button entity instead"
         )
         for coordinator, obj in iter_objects(service_call, mustbe=Charger):
             _LOGGER.debug("  >> to %s", obj.id)
--- a/custom_components/zaptec/services.py
+++ b/custom_components/zaptec/services.py
@@ -185,41 +185,25 @@ async def async_setup_services(hass: HomeAssistant, manager: ZaptecManager) -> N
             else:
                 err_device = f"id {uid}"
 
-            # Get all coordinators and objects that matches the uid from all coordinator objects
-            matches = set(
-                (manager.coordinator, obj)
-                for obj in manager.zaptec.objects()
-                if obj.id == uid
-            )
-            # Filter out the objects that doesn't match the expected type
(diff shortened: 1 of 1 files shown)


custom_components/zaptec/services.py
```diff
@@ -185,41 +185,25 @@ def iter_objects(
...
-            # Send to caller
-            yield from want
+            coordinator = manager.device_coordinators[uid]
+            yield coordinator, zaptec_object
 
     async def service_handle_stop_charging(service_call: ServiceCall) -> None:
         _LOGGER.debug("Called stop charging")
+        _LOGGER.warning(
+            "The 'stop_charging' action is deprecated and will be removed in a future release. "
+            "Use the 'stop_charging' botton entity instead"
```
steinmn: misspelled 'button', and all the case for the button names in the warnings should be consistent
```suggestion
            "Use the 'Stop charging' button entity instead"
```

custom_components/zaptec/services.py
```diff
@@ -228,10 +212,14 @@ async def service_handle_stop_charging(service_call: ServiceCall) -> None:
...
                     f"Command 'stop_charging_final' failed: {exc}"
                 ) from exc
-            await coordinator.trigger_poll(obj)
+            await coordinator.trigger_poll()
 
     async def service_handle_resume_charging(service_call: ServiceCall) -> None:
         _LOGGER.debug("Called resume charging")
+        _LOGGER.warning(
+            "The 'resume_charging' action is deprecated and will be removed in a future release. "
+            "Use the 'Resume Charging' button entity instead"
```
steinmn:
```suggestion
            "Use the 'Resume charging' button entity instead"
```

custom_components/zaptec/services.py
```diff
@@ -307,7 +311,7 @@ async def service_handle_limit_current(service_call: ServiceCall) -> None:
                 )
             except Exception as exc:
                 raise HomeAssistantError(f"Limit current failed: {exc}") from exc
-            await coordinator.trigger_poll(obj)
+            await coordinator.trigger_poll()
 
     async def service_handle_send_command(service_call: ServiceCall) -> None:
         _LOGGER.debug("Called send command")
```
steinmn, line 317: ~~Should we include an is_command_valid-check on this?̃~~

steinmn (commented):
Minor inconsistencies on the warning messages, ~~and I think we should catch invalid commands before calling the api~~

custom_components/zaptec/services.py
steinmn: :facepalm: Sorry, just realized we built that into the command-function, nevermind

sveinse: Yeah I was looking at that `iter_objects()` function yesterday and was going WTF. It definitely can be made simpler. Its hard to understand as it it, but at this point I don't want to open that box of worms. Goal now is to get v0.8 out.

steinmn (approved):
LGTM

PR #245, opened 2025-07-25
Replace unavailable common string-keys with the translated strings

### LINKED ISSUE #242: Missing translations on wrong password
Incorrect password on configuration is not properly translated. Instead it prints

`[%key:common::config_flow::error::invalid_auth%]`

Might there be any other "standard" texts that we are missing?

steinmn:
Only core components can use these standard texts: https://developers.home-assistant.io/docs/internationalization/custom_integration/

(translations collected from [airnow component last commit before translations were removed from core](https://github.com/home-assistant/core/tree/4ad386d7941ba0bc6cec7674f3fea85357194f8b/homeassistant/components/airnow/translations) )

Fixes #242

### DIFF
--- a/custom_components/zaptec/translations/en.json
+++ b/custom_components/zaptec/translations/en.json
@@ -23,8 +23,8 @@
         "error": {
             "cannot_connect": "Failed to connect",
             "invalid_auth": "Invalid authentication",
-            "unknown": "Unexpected error",
-            "no_chargers_selected": "No chargers selected. Please select at least one charger."
+            "no_chargers_selected": "No chargers selected. Please select at least one charger.",
+            "unknown": "Unexpected error"
         },
         "abort":{
             "already_exists": "One instance of Zaptec already exists."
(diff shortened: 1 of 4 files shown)


custom_components/zaptec/translations/en.json
```diff
@@ -21,9 +21,9 @@
             }
         },
         "error": {
-            "cannot_connect": "[%key:common::config_flow::error::cannot_connect%]",
-            "invalid_auth": "[%key:common::config_flow::error::invalid_auth%]",
-            "unknown": "[%key:common::config_flow::error::unknown%]",
+            "cannot_connect": "Failed to connect",
+            "invalid_auth": "Invalid authentication",
+            "unknown": "Unexpected error",
```
sveinse: Another HA check thing that's coming is that this list must be in alphabetical order. Since we're touching these lines, please swap the order of "unknown" and "no_chargers_selected" swap place.

sveinse (commented):
It looks good. Get the swap key order in place, and we're good to go.

custom_components/zaptec/translations/en.json
steinmn: Done

PR #247, opened 2025-07-26
Updated ruff settings and fixed the simplest errors in the code

sveinse:
This PR fixes the most low-hanging fruits of the lint errors in the code. The PR also configures a few exceptions to ruff, that are too strict. I've used HA as a template to guide what settings should be enabled or disabled.

There are more lint errors remaining, and when someone takes a stab at it, we should also continue to evaluate if the rule is valid or not. We should keep looking at the extensive HA setting to see if we should mirror HAs exception.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,18 +1,27 @@
 # The contents of this file is based on https://github.com/home-assistant/core/blob/dev/pyproject.toml
 
 target-version = "py313"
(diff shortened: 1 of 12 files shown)


steinmn (commented):
Running `ruff format .` also adds updates for number.py and update.py, please include those as well.

Otherwise LGTM

PR #250, opened 2025-07-27
Add support for adding standalone chargers

### LINKED ISSUE #241: Fails to add chargers when access denied to installation
It was assumed in v0.8 that all chargers have an integration and that we ignore the standalone chargers (i.e. those without installation). However, testing, initiated by @TorbjornMohn reveals that this is not the case. If the user doesn't have access to the installation, then the chargers won't get added.

I am able to repro this using a limited test user for my own installation.

```
2025-07-25 19:47:52.840 WARNING (MainThread) [custom_components.zaptec.api] Access denied to installation hierarchy of {id}. The user might not have access
2025-07-25 19:47:52.890 WARNING (MainThread) [custom_components.zaptec.api] These standalone chargers will not be added: {'{id}'}
```

sveinse:
* This is needed when user doesn't have access to the installation

Fix #241

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import AsyncGenerator, Callable, Iterable, Iterator, Mapping
 from contextlib import aclosing
+import itertools
 import json
 import logging
 import random
@@ -284,7 +285,7 @@ class Installation(ZaptecBase):
                     charger: Charger = self.zaptec[charger_item["Id"]]
                     charger.set_attributes(charger_item)
                 else:
-                    _LOGGER.debug("      Charger %s", charger_item["Id"])
+                    _LOGGER.debug("      Charger %s  (adding)", charger_item["Id"])
                     charger = Charger(charger_item, self.zaptec, installation=self)
                     self.zaptec.register(charger_item["Id"], charger)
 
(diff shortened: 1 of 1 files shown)


custom_components/zaptec/api.py
```diff
@@ -1209,10 +1210,41 @@ async def build(self):
...
+        )
+
+        # Add standalone chargers that are not part of any installation.
+        # Users without service access right does not have access to the installation
+        # object, so we need to add all the object at this point.
+        for charger_item in chargers["Data"]:
+            if charger_item["Id"] in installation_chargers:
+                continue  # Skip the chargers which have already been found in installations
+            if charger_item["Id"] in self:
+                _LOGGER.debug("  Charger %s  (existing)", charger_item["Id"])
```
```suggestion
                _LOGGER.debug("  Standalone Charger %s  (existing)", charger_item["Id"])
```

custom_components/zaptec/api.py
```diff
@@ -1209,10 +1210,41 @@ async def build(self):
...
+                charger: Charger = self[charger_item["Id"]]
+                charger.set_attributes(charger_item)
+            else:
+                _LOGGER.debug("  Charger %s  (adding)", charger_item["Id"])
```
```suggestion
                _LOGGER.debug("  Standalone Charger %s  (adding)", charger_item["Id"])
```

custom_components/zaptec/api.py
sveinse: Both fixed. Thanks.

PR #252, opened 2025-07-27
Release v0.8b2

sveinse:


steinmn:
Ran a full round with my sun-charging automation on 0.8b2 today, and the only thing I noticed was that the values appeared to react much quicker than before. Then again I'm sliiiiightly biased here. 

My experience with length of release process ranges from 1,5 months to 3+ months, but that SW was on a completely different scale than this. The only major issue was a misjudgement  on our side of the usecase for standalone chargers, wasn't it? And that was quite contained. The biggest change/ I would say was the change in polling architecture, and I've seen no issues from that. So maybe give it until the end of the week?

Quick question since you've used v0.8b# consistently: The final version will still be v0.8.0, right? Not v0.8?

sveinse:
> Quick question since you've used v0.8b# consistently: The final version will still be v0.8.0, right? Not v0.8?

I was thinking `v0.7.4 -> v0.8b* -> v0.8 -> v0.8.1b* -> v0.8.1`. I certainly don't like long `1.0.2.34` versions. But perhaps `v0.8.0` is better`, when I see this sequence.

steinmn:
All releases up until now have been 3 numbers (including v0.7.0), so I think it's best to stick to the pattern.

PR #254, opened 2025-07-28
Minor improvements

sveinse:
* Use `obj.qual_id` instead of `obj.id` for logging, due to not log the full UID
* Improve diagnostics functions. Use more natuaral language in variables
* Improve and refactor Redactor
* Some optimizations in api

### DIFF
--- a/custom_components/zaptec/diagnostics.py
+++ b/custom_components/zaptec/diagnostics.py
@@ -101,7 +101,7 @@ class Redactor:
         self.redact_info = {}
 
     def dumps(self) -> str:
-        """Dump the redation database in a readable format."""
+        """Dump the redaction database in a readable format."""
         return pformat(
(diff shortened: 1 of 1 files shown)


custom_components/zaptec/diagnostics.py
```diff
@@ -91,24 +94,34 @@ class Redactor:
...
+    def __init__(self, do_redact: bool, obs_ids: dict[str, str] | None = None):
         self.do_redact = do_redact
-        self.obs_ids = obs_ids
+        self.obs_ids = obs_ids or {}
         self.redacts = {}
         self.redact_info = {}
 
-    def add_redact(self, obj, ctx=None, key=None, redact=None) -> str:
+    def dumps(self) -> str:
+        """Dump the redation database in a readable format."""
```
steinmn:
```suggestion
        """Dump the redaction database in a readable format."""
```

PR #256, opened 2025-07-28
Feature add log redaction

### LINKED ISSUE #255: Prevent sharing of sensitive data in debug logs
Often when wanting to help users with debugging why Zaptec isn't working for them, access to the Zaptec HA detailed debug logs are often required.  However, the Zaptec API library is logging a lot of sensitive information to the logs. By sensitive, I mean information that should not be put out publicly on the net, such as in an issue here. These are things like names, UIDs, addresses and position. Diagnostics is providing a system for redacting the information, but there is none for logs.

After conferring with other HA devs on Discord, it seems that its considered an anti-pattern to log sensitive information to the logs. Despite this, many critical libraries will do that if one enable debug. E.g. enabling debug on sshd will put a lot of sensitive stuff in the logs.[^1]

The questions raised in this issue: 
1. How can we get the logs from the user needing help without disclosing any sensitive information?
2. How can we prevent sensitive data in the logs while still having the detailed debug logs?

[^1]: HA integrations are normally divided into two libraries: (1) _access library_ which communicates to the remote service, and (2) the HA integration itself. When talking about debug mode in HA, its normally only specifying the HA integration, not the access library. Our Zaptec integration has both, and its the access library part that "leaks" most into logs.

sveinse:
This is a proposal to fix #255. The PR adds a system for redaction of sensitive data. This is the same system previously used for the diagnostics download, which is now repurposed for all logging from api.py. The overall idea is that it should be safe to share the debug logs -- with the exception of the redaction database which is clearly marked in the log.

A log entry typically looks like this with all sensitive data redacted:

```
@@@  REQUEST GET to 'https://api.zaptec.com/api/installation'
@@@  RESPONSE 200 length 1118
   Installation <--Inst[8a106b]-->  (adding)
>>>  Adding   Installation.id (Id)  =  <str> '<--Inst[8a106b]-->'
>>>  Adding   Installation[8a106b].name (Name)  =  <str> '<--Redact #2-->'
>>>  Adding   Installation[8a106b].address (Address)  =  <str> '<--Redact #3-->'
```

The redacted stings are found in `"<--Text-->"` brackets. At the end of setup if `REDACT_DUMP_ON_STARTUP` is set, will print the following in the logs.

```
------  DO NOT PUBLISH THE FOLLOWING DATA  ------
|  Redaction database:
|  {'<--Inst[8a106b]-->': '...the actual value....',
|   '<--Redact #2-->': '...the actual value....',
|   '<--Redact #3-->': '...the actual value....'}
------  DO NOT PUBLISH THE ABOVE ^^^ DATA  ------
```

The setting `REDACT_LOGS = False` (in const.py) can be used to disable redaction completely. I've chosen to set `REDACT_DUMP_ON_STARTUP = True` so there is a chance to extract the actual values from the logs if need be. But users must make sure they don't share that part of the logs.

The downside to this solution, is that it adds even more complexity to the library and that we can never guarantee 100% that any sensitive information will not fall through the filter.

Is this PR a good idea, or just added complexity?

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -43,6 +43,8 @@ from .const import (
     KEYS_TO_SKIP_ENTITY_AVAILABILITY_CHECK,
     MANUFACTURER,
     MISSING,
+    REDACT_DUMP_ON_STARTUP,
+    REDACT_LOGS,
     REQUEST_REFRESH_DELAY,
     ZAPTEC_POLL_CHARGER_TRIGGER_DELAYS,
     ZAPTEC_POLL_INSTALLATION_TRIGGER_DELAYS,
@@ -123,6 +125,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         client=async_get_clientsession(hass),
         max_time=ZAPTEC_POLL_INTERVAL_CHARGING,  # The shortest of the intervals
         show_all_updates=True,  # During setup we'd like to log all updates
+        redact_logs=REDACT_LOGS,
     )
 
     # Login to the Zaptec account
@@ -223,6 +226,14 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     # will create a lot of debug log output.
     zaptec.show_all_updates = False
     _LOGGER.debug("Zaptec setup complete")
+
+    # Dump the redaction database
+    if REDACT_LOGS and REDACT_DUMP_ON_STARTUP:
+        _LOGGER.debug("------  DO NOT PUBLISH THE FOLLOWING DATA  ------")
+        _LOGGER.debug("|  Redaction database:")
+        for line in manager.zaptec.redact.dumps().splitlines():
+            _LOGGER.debug("|  %s", line)
+        _LOGGER.debug("------  DO NOT PUBLISH THE ABOVE ^^^ DATA  ------")
 
     # Attach the local data to the HA config entry so it can be accessed later
     # in various HA functions.
--- a/custom_components/zaptec/redact.py
+++ b/custom_components/zaptec/redact.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from pprint import pformat
 from typing import ClassVar, TypeVar, cast
 
 T = TypeVar("T")
--- a/custom_components/zaptec/redact.py
+++ b/custom_components/zaptec/redact.py
@@ -1,0 +1,166 @@
+"""Helper class to redact sensitive data from the API logging."""
+
+from __future__ import annotations
+
+from typing import ClassVar, TypeVar, cast
+
+T = TypeVar("T")
+
+
+class Redactor:
+    """Class to handle redaction of sensitive data."""
+
+    # Data fields that must be redacted from the output
+    REDACT_KEYS: ClassVar[list[str]] = [
+        "Address",
+        "ChargerId",
+        "ChargerCurrentUserUuid",
+        "CircuitId",
+        "City",
+        "DeviceId",
+        "Id",
+        "ID",
+        "InstallationId",
+        "InstallationName",
+        "Latitude",
+        "LogoBase64",
+        "Longitude",
+        "LteIccid",
+        "LteImei",
+        "LteImsi",
+        "MacWiFi",
+        "MacMain",
+        "MacPlcModuleGrid",
+        "MID",
+        "Name",
+        "NewChargeCard",
+        "PilotTestResults",
+        "Pin",
+        "ProductionTestResults",
+        "SerialNo",
+        "SupportGroup",
+        "ZipCode",
+    ]
+
+    # Never redact these words
+    NEVER_REDACT: ClassVar[list] = [
+        None,
+        True,
+        False,
+        "true",
+        "false",
+        "0",
+        "0.",
+        "0.0",
+        0,
+        0.0,
+        "1",
+        "1.",
+        "1.0",
+        1,
+        1.0,
+        "",
+    ]
+
+    # Keys that will be looked up into the observer id dict
+    OBS_KEYS: ClassVar[list[str]] = ["SettingId", "StateId"]
+
+    # Key names that will be redacted if they the dict has a OBS_KEY entry
+    # and it is in the REDACT_KEYS list.
+    VALUES: ClassVar[list[str]] = [
+        "Value",
+        "ValueAsString",
+    ]
+
+    def __init__(self, do_redact: bool, obs_ids: dict[str, str] | None = None):
+        self.do_redact = do_redact
+        self.obs_ids = obs_ids or {}
+        self.redacts = {}
+        self.redact_info = {}
+
+    def dumps(self) -> str:
+        """Dump the redaction database in a readable format."""
+        return pformat(
+            {k: v["text"] for k, v in self.redact_info.items()},
+        )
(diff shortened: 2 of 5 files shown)


custom_components/zaptec/redact.py
```diff
@@ -0,0 +1,167 @@
...
+                    for k, v in obj.items()
+                },
+            )
+
+        # Check if the object is already redacted
+        if obj in self.redacts:
+            return self.redacts[obj]
+
+        # Check if new redaction is needed
+        if key and key in self.REDACT_KEYS and (not obj or obj not in self.NEVER_REDACT):
```
steinmn: What is the purpose of `not obj` here?

In my setup, latitude and longitude has not been set, leading them to become 0.0. I think that since `not 0.0` is true/truthy, that overrides the NEVER_REDACT.  At least latitude is the first instance I found of `<--Redact #4-->`, and all subsequent zero-floats in my debug logs are also redacted.

custom_components/zaptec/redact.py
```diff
@@ -0,0 +1,167 @@
...
+        "LogoBase64",
+        "Longitude",
+        "LteIccid",
+        "LteImei",
+        "LteImsi",
+        "MacWiFi",
+        "MacMain",
+        "MacPlcModuleGrid",
+        "MID",
+        "Name",
```
steinmn, line 36: All of the entites are autogenerated with based on the name of the device, aren't they? So for instance the entity_id of "Charger mode" is `<charger_name_in_snake_case>_charger_mode`. If we redact the name, don't we also need to redact the name from the entity_id as well?

custom_components/zaptec/__init__.py
```diff
@@ -224,6 +227,14 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     zaptec.show_all_updates = False
     _LOGGER.debug("Zaptec setup complete")
 
+    # Dump the redaction database
+    if REDACT_LOGS and REDACT_DUMP_ON_STARTUP:
+        _LOGGER.debug("------  DO NOT PUBLISH THE FOLLOWING DATA  ------")
+        _LOGGER.debug("|  Redaction database:")
+        for line in manager.zaptec.redact.dumps().splitlines():
+            _LOGGER.debug("|  %s", line)
+        _LOGGER.debug("------  DO NOT PUBLISH THE ABOVE ^^^ DATA  ------")
```
steinmn: This is in an async method, can we trust that there won't be other lines inserted inbetween these? I think generating the full string first and then logging it in a single debug-call would be preferable.

custom_components/zaptec/redact.py
sveinse: Yes, the `or` is incorrect here: It should read
```py
if key and key in self.REDACT_KEYS and objand obj not in self.NEVER_REDACT:
```

custom_components/zaptec/redact.py
sveinse: It is not easy to select what to redact or not for that very precise reason. By default chargers use their `DeviceId` such as `ZAP123456` which is definitely private information. Some users give them their personal name, which is definitely sensitive. So the naming of entities part is currently not redacted.

If the user gives this a very common name, say "0", then _every_ use of that string will be redacted from `api.py`, which is very unfortunate. Like you experienced with the 0-redaction.

custom_components/zaptec/__init__.py
sveinse: Yes, this is an async function, and these are not async, so they should be atomic. Unless HA uses multithreaded asyncio, which is very exotic, I believe this will be logged together. I know HA does a lot of unorthodox trickeries, so I'll ask around.

Logging with newlines are frowned upon. However, they don't break up TB to ensure it stays together, so perhaps this is a single log entry.

custom_components/zaptec/redact.py
sveinse: Fixed, please retest. I didn't have any `0.0` items in my setup, so I missed this part.

This leads me into another topic we should also add on our issue list: Making a test suite. That should avoid mistakes like this. I've started a few times on the framework for pytest, but haven't really been able to commit fully. It does require some work getting up.

custom_components/zaptec/__init__.py
sveinse: I changed this to log into one operation.

custom_components/zaptec/redact.py
steinmn: Hmmm, yeah, that one is a bit tricky. Maybe add a point in the readme-entry on [debugging](https://github.com/custom-components/zaptec?tab=readme-ov-file#debugging) to specifically be on the lookout for this and use search-and-replace?

custom_components/zaptec/redact.py
steinmn: Looks good

Yeah, tests are good to have, and getting the framework for it up and running is often the biggest hurdle.

steinmn:
> The risk of ending up with redacting common words like "0.0" is much higher.

Yeah, let's not tempt fate with that.

I think the code is ok as it is now, just do a revision of the readme and maybe add some instructions to development.md

custom_components/zaptec/redact.py
sveinse: I asked around in HA discord and the entity names isn't normally considered sensitive. So I think we're good at redacting in the API access library only (as we have), and let the HA facing parts be as it is.

I agree we should warn the user in the docs that the redaction is mainly to prevent accidental data leakage, not to be a water tight filter. The user must always review the data before the data is shared or made public.

README.md
```diff
@@ -306,16 +306,34 @@ logger:
...
-> filtering them.
-
 > [!NOTE]
 > The Zaptec integration logs massive amounts in
 > debug. This is nice for finding errors, but it will generate large amount of
 > data if left enabled for long. Do not use in production setups.
 
+#### Redaction filter
+
+The integration have by default a *redaction filter* which replaces sensitive
```
steinmn:
```suggestion
The integration has a *redaction filter* which by default replaces sensitive
```

README.md
```diff
@@ -306,16 +306,34 @@ logger:
...
+
+#### Sharing debug logs
+
+> [!WARNING]
+> The debug logs may contain identifiable information about your Zaptec setup
+> such as name or serial numbers. The redaction filter is designed to remove
+> most of sensitive information, but it is not perfect. Please review any
+> logs before it is shared or made public.
+>
+> The Zaptec entity names in might contain sensitive names. These are not
```
steinmn:
```suggestion
> The Zaptec entity names might contain sensitive information. These are not
```

DEVELOPMENT.md
```diff
@@ -32,6 +32,27 @@ should be inclined to follow suit.
 - https://github.com/home-assistant/core/blob/dev/pyproject.toml#L652
 
 
+## Logging
+
+The zaptec integration have a number of internal ways to increase debug logging
```
steinmn:
```suggestion
The Zaptec integration has a number of internal ways to increase debug logging
```

steinmn (approved):
Just a couple of typos, LGTM

sveinse:
@steinmn Thank you for the spelling corrections. I probably need to get myself a contextual spell checker in VS Code 😁

sveinse:
Thank you once again @steinm for helping out.

sveinse:
@steinmn do we want this and #254 in or out of v0.8.0 ?

steinmn:
My vote is to include them👍

Issue #258, opened 2025-07-29
Fix all ruff-issues and enable automatic linting


steinmn:
Automatic linting is currently disabled because there are over 300 code quality issues that need to be resolved before the test would pass.

sveinse:
Code quality created. The function "sub-issue" is also good thing to catch them under one (this?) umbrella.

sveinse:
When looking at the lint errors ruff provides, I think we shouldn't blindly start changing to code to pass the lint error. We also need to evaluate if that particular rule makes sense for us. HA has a very long list of exceptions and we've just adopted a few. So we should also update the exception rules where it makes sense.

rhammen:
Progress note: **PR #390** clears all ruff errors in `custom_components/zaptec/zaptec/api.py` (36 → 0) — api.py held the bulk of the remaining lint debt. Once the rest of the tree is clean, the CI `ruff check` step (currently run with `--exit-zero`) can be switched to enforcing so this stays green.

Offline unit-test coverage for api.py landed alongside in **PR #391** (~13% → ~66%), which makes further cleanup safer to verify.

sveinse:
Once we've merged these two PRs, should we consider this issue as closed? If so, we can label this for the v0.8.8 milestone.

rhammen:
I'd say we can close this issue when these 2 PRs are merged.

PR #261, opened 2025-07-29
Add more explicit typing

steinmn:
`ruff check .` now down to 308 errors. part of #258

### DIFF
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -482,7 +482,7 @@ class ZaptecManager:
 
         return entities
 
-    def create_streams(self):
+    def create_streams(self) -> None:
         """Create the streams for all installations."""
         for install in self.zaptec.installations:
             if install.id in self.zaptec:
@@ -496,7 +496,7 @@ class ZaptecManager:
                 )
                 self.streams.append((task, install))
 
-    async def cancel_streams(self):
+    async def cancel_streams(self) -> None:
         """Cancel all streams for the account."""
         for task, install in self.streams:
             _LOGGER.debug("Cancelling stream for %s", install.qual_id)
@@ -702,7 +702,7 @@ class ZaptecUpdateCoordinator(DataUpdateCoordinator[None]):
             finally:
                 self._trigger_task = None
 
-        def cleanup_task(_task: asyncio.Task):
+        def cleanup_task(_task: asyncio.Task) -> None:
             """Cleanup the task after it has run."""
             self._trigger_task = None
 
@@ -792,6 +792,7 @@ class ZaptecBaseEntity(CoordinatorEntity[ZaptecUpdateCoordinator]):
         Raises:
             KeyUnavailableError: If key doesn't exist or obj doesn't have
             `.get()`, which indicates that obj isn't a Mapping-like object
+
         """
         obj = self.zaptec_obj
         key = key or self.key
@@ -830,7 +831,7 @@ class ZaptecBaseEntity(CoordinatorEntity[ZaptecUpdateCoordinator]):
         return f".{v}"
 
     @callback
-    def _log_value(self, attribute: str | None, force=False):
+    def _log_value(self, attribute: str | None, force=False) -> None:
         """Helper to log a new value."""
         if attribute is None:
             return
@@ -849,7 +850,7 @@ class ZaptecBaseEntity(CoordinatorEntity[ZaptecUpdateCoordinator]):
             )
 
     @callback
-    def _log_unavailable(self, exception: Exception | None = None):
+    def _log_unavailable(self, exception: Exception | None = None) -> None:
         """Helper to log when unavailable."""
         available = self._attr_available
         prev_available = self._prev_available
@@ -880,7 +881,7 @@ class ZaptecBaseEntity(CoordinatorEntity[ZaptecUpdateCoordinator]):
             _LOGGER.info("Entity %s is available", self.entity_id)
 
     @property
-    def key(self):
+    def key(self) -> str:
         """Helper to retrieve the key from the entity description."""
         return self.entity_description.key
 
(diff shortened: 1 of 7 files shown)


custom_components/zaptec/zconst.py
```diff
@@ -95,80 +98,79 @@ def update_ids_from_schema(self, device_types):
...
 
-    def type_user_roles(self, v):
+    def type_user_roles(self, val: int) -> str:
         """Convert the user roles to a string."""
-        val = int(v)
-        if not val:
+        if val == 0:
             return "None"
-        roles = set(k for k, v in self.get("UserRoles", {}).items() if v & val == v)
+        roles = {k for k, v in self.get("UserRoles", {}).items() if v != 0 and v & val == v}
```
steinmn: Note: added `v != 0` because
```Current user roles        User, Owner, None```
looks a bit silly

File: custom_components/zaptec/__init__.py [hunk 2]
```diff
@@ -792,6 +792,7 @@ def _get_zaptec_value(
         Raises:
             KeyUnavailableError: If key doesn't exist or obj doesn't have
             `.get()`, which indicates that obj isn't a Mapping-like object
+
```
sveinse, line 800: Nitpick, did you add this, or was this the auto-formatter? Did you run autoformat on any of these?

steinmn:
@sveinse I took a pass and handled some of the more obvious/simpler ruff-complaints. Up to you if you want to look at this now, or leave it for after 0.8.0 is released.

custom_components/zaptec/zconst.py
```diff
@@ -43,7 +44,9 @@ class ZConst(UserDict):
     ]
     """Valid parameters for charger settings (`chargers/{id}/update`)."""
 
-    def get_remap(self, wanted, device_types=None) -> dict:
+    def get_remap(
+        self, wanted: list[str], device_types: Iterable[str] | None = None
```
sveinse: I wonder if `Iterable[str]` is not the right type here, as the "in" operation is used and `Iterable` doesn't have `__contains__`. Interesting that the linter doesn't react to that.

See https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

custom_components/zaptec/zconst.py
```diff
@@ -95,80 +98,79 @@ def update_ids_from_schema(self, device_types):
...
+    def type_ocmf(self, data: str) -> dict[str, Any]:
         """Open Charge Metering Format (OCMF) type."""
         # https://github.com/SAFE-eV/OCMF-Open-Charge-Metering-Format/blob/master/OCMF-en.md
         sects = data.split("|")
         if len(sects) not in (2, 3) or sects[0] != "OCMF":
-            raise ValueError(f"Invalid OCMF data: {data}")
-        data = json.loads(sects[1])
-        return data
+            msg = f"Invalid OCMF data: {data}"
+            raise ValueError(msg)
```
sveinse: This is one of the linting error that I'm _really_ uncertain about. This is the first time I've heard about this rule via ruff, and I'm not seeing HA adhering to this.

custom_components/zaptec/zconst.py
sveinse: I don't quite understand this one. Isn't `0` falsey and the ig `if v` would filter it out? There is something here I'm not getting.

steinmn: Maybe we should add extra parentheses for clarity?
`if (v != 0) and (v & val) == v`
The problem is that `v & val` (bitwise and) evaluates to 0 when v=0, and `0 == 0` is true.

sveinse: Oh, indeed. That was a sneaky one. I suggest adding a comment above saying that the v != 0 is to prevent 0 == 0.

custom_components/zaptec/__init__.py
steinmn: I did run `ruff check . --fix` once (by accident, had forgotten to remove --fix), but I removed the ones that I didn't feel were obvious keepers.

This one is https://docs.astral.sh/ruff/rules/missing-blank-line-after-last-section/#missing-blank-line-after-last-section-d413, and while it doesn't look like many components have gotten around to adding `Raises:` to their docstrings, the ones who have have included the last blank line https://github.com/search?q=repo%3Ahome-assistant%2Fcore%20Raises%3A&type=code

custom_components/zaptec/zconst.py
steinmn: Hmm looks like I've tried to be too general, maybe just go for `set` instead?

custom_components/zaptec/zconst.py
steinmn: Ok, I'll remove it here, and we leave it for later

custom_components/zaptec/zconst.py
sveinse: `set`, `list` and `dict` have `Container[str]` in common

steinmn: Do you want it changed to `Container`, or is it ok with `set`? We only call the function in one location, and it is called on a set 🤷

sveinse: I think either is ok. If you chose `set` and if someone later wants to extend this to a more generic type, then we can make the more generic change then. -- But what we do want to avoid later is someone doing `.get_remap(wanted, device_type=set(a_list))` to satisfy the linter.

custom_components/zaptec/__init__.py
sveinse: I'm happy, I was just curious.

custom_components/zaptec/zconst.py
steinmn: Ok, then I think I'll leave it as set.

steinmn:
Done

sveinse:
Merged. Thank you.

PR #264, opened 2025-07-30
Add new "3 to 1-phase switch current" number in installation

### LINKED ISSUE #172: Zaptec Go2 Changing between 1 and 3-phase charging
I've got a new Zaptec Go 2 Wallbox I'm trying to configure for PV charging. To use most of the surplus solar power, I need to change between 1-phase and 3-phase charging. I can do this in the Zaptec app but cannot find the option in the integration.

sveinse:
Fixes #172 

I do not have a system that support this function, so I am not able to any type of testing and verification of this. It would be nice if someone could chip in here.

* Add new number entity class `ZaptecThreeToOnePhaseSwitchCurrent()`
* Add `Installation.set_three_to_one_switch_current()` method

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ the same time.
 > current. Zaptec recommends not changing the values more often than 15 minutes.
 
 
-#### 3 phase current adjustment
+### 3 phase current adjustment
 
 The service call `limit_current` can be used with the arguments
 `available_current_phase1`, `available_current_phase2` and


Revilo-S:
Thanks a lot for your work.

I can confirm the behaviour.
"By changing to 3to1 current setting to 32 I can force 1-phase and by setting it to 0 I can force 3-phase"
Still, I think it would be nicer to have a simple switch that changes between 1- and 3-phase charging.

Revilo-S:
Then I'll have to do a little more thinking myself when I'm implementing my charge control with nodered. 
I guess it'll be no problem.

Thanks again.

custom_components/zaptec/number.py
```diff
@@ -36,6 +37,17 @@ def _update_from_zaptec(self) -> None:
         self._attr_native_value = self._get_zaptec_value()
         self._attr_available = True
 
+    def _log_number(self, value: Any) -> None:
```
steinmn:
```suggestion
    def _log_number(self, value: float) -> None:
```

custom_components/zaptec/number.py
```diff
@@ -4,6 +4,7 @@
 
 from dataclasses import dataclass, replace
 import logging
+from typing import Any
```
steinmn: not needed if log_number uses float
```suggestion
```

sveinse:
> At some point I think it would be nice to avoid adding the `three_to_one_phase_switch_current`-entity for setups that don't support it anyway But since it's a little bit weird with the setting on the installation level and the model on the charger level, that can quickly get complicated, so I think we'll leave that as a future feature request.

Zaptec did tell us that they have future plans for "fixing" this somewhat strange placement in future updates. So I think we should wait until they add this feature in the right place. See https://github.com/custom-components/zaptec/issues/172#issuecomment-3136076835

> I still think it would be a good idea to add a reminder in the readme that this setting will not do anything if you don't have the correct charger model, so that if/when we get bug reports about it not working, we can point to the explanation there.

Initially I had plans on just adding this entity for the chargers that supports it. But since this is a installation thing and not a charger thing, its not that easy. And while testing this on my system, which does not support phase switching, my installation now started reporting `three_to_one_phase_switch_current`. See https://github.com/custom-components/zaptec/issues/172#issuecomment-3136458752

My thinking with this integration is that we're just a transparent proxy between HA and Zaptec Portal. We should not be stateful, nor should we interpret or filter the values. And we should not be documentation of how to use Zaptec. If Zaptec Portal presents `three_to_one_phase_switch_current`, we should show it. Regardless if the chargers support it or not.

custom_components/zaptec/number.py
sveinse: I was really uncertain if I agreed to this. The function actually does not require value to be float, it can be anything. However the name and its usage seems to suggest so.

I fixed as suggested. For this usage, I agree this is best.

sveinse:
Added README documentation section about this.

custom_components/zaptec/number.py
steinmn: Yes, if the function for instance had been named `_log_value`, I probably could have been persuaded to accept Any.

sveinse: You see, `_log_value()` is already used the base class for a slightly different purpose. That's the reason for it's somewhat strange name :D

README.md
```diff
@@ -225,6 +225,28 @@ The service call `limit_current` can be used with the arguments
 `available_current_phase3` to set the available current on individual phases.
 
 
+### 3 phase to 1 phase switch
```
steinmn, line 228: (nitpick) I think the `3 phase current adjustment`-headline above this one should also have ### for consistency, otherwise it goes 2-4-3 in the headline leveling

steinmn (approved):
LGTM

README.md
sveinse: Fixed

sveinse:
@steinmn Thank you very much

PR #265, opened 2025-07-30
Minor improvements

sveinse:
While working on other issues, these two were discovered and improved.

* Change text when entity is not available
* Add checks of the values for `Installation.set_limit_current`.

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -546,13 +546,22 @@ class Installation(ZaptecBase):
                 "availableCurrentPhase3",
             )
         )
-
         if not (has_availablecurrent ^ has_availablecurrentphases):
             raise ValueError(
                 "Either availableCurrent or all of availableCurrentPhase1, "
                 "availableCurrentPhase2, availableCurrentPhase3 must be set"
             )
-
+        # Make sure the arguments and values are valid
+        for k, v in kwargs.items():
+            if k not in (
+                "availableCurrent",
+                "availableCurrentPhase1",
+                "availableCurrentPhase2",
+                "availableCurrentPhase3",
+            ):
+                raise Error(f"Invalid argument {k!r}")
+            if not (0 <= v <= 32):
+                raise ValueError(f"{k} must be between 0 and 32 amps")
         data = await self.zaptec.request(
             f"installation/{self.id}/update", method="post", data=kwargs
         )
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -551,6 +551,11 @@ class Installation(ZaptecBase):
                 "Either availableCurrent or all of availableCurrentPhase1, "
                 "availableCurrentPhase2, availableCurrentPhase3 must be set"
             )
+        # Use 32 as default if missing or invalid value.
+        try:
+            max_current = float(self.get("max_current", 32.0))
+        except (TypeError, ValueError):
+            max_current = 32.0
         # Make sure the arguments and values are valid
         for k, v in kwargs.items():
             if k not in (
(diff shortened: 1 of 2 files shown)


sveinse:
We can create a master-next if we want to proceed with these merges now. If for some reason we need a `v0.8b3` we can chose to merge them into master. If not we'll do it after v0.8.0 is out.

custom_components/zaptec/api.py
```diff
@@ -546,13 +546,22 @@ async def set_limit_current(self, **kwargs):
...
+        for k, v in kwargs.items():
+            if k not in (
+                "availableCurrent",
+                "availableCurrentPhase1",
+                "availableCurrentPhase2",
+                "availableCurrentPhase3",
+            ):
+                raise Error(f"Invalid argument {k!r}")
+            if not (0 <= v <= 32):
+                raise ValueError(f"{k} must be between 0 and 32 amps")
```
steinmn: Shouldn't the max be determined by the max_current of the installation? While I would expect 32 to be the most common, both less and more is possible.

sveinse: Because the guide from Zaptec sais so: https://docs.zaptec.com/docs/3-to-1-phase-switching-with-zaptec-go-2#controlling-phase-switching-behavior-with-the-zaptec-api

sveinse: This number dial does not explain how to use it, so the user must be aware of the significance of setting the value to 0, 1-31 and 32.

I'm not sure there is any good way to handle the GUI for this in HA without creating a lot of logic. In Zaptec Portal, this is a simple number input, so having a number dial is no different from that.

steinmn: But this isn't the switching function, this is setting the available current. As far as I can see, the linked documentation is only referring to phaseswitching behaviour.

sveinse: Oh, I'm sorry, I mixed this up with the other PR.

Yeah, we could compare it against the reported max current for that installation for the upper limit, yes.

steinmn: I'd really like this to also not allow the values in between zero and min_charger_current (while still allowing zero), but since min_charger_current is on the charger level and this is on the installation level, it could get quite messy. Any ideas? Is the don't-go-below-6A-rule general enough that we could hardcode it?

sveinse: Have no idea. Need to ask Zaptec that. Zaptec Portal lets me set "2" and are happy with that, so they don't enforce any limits.

How would you calculate the minimum permittable value? `min(charger["charger_min_current"] for charger in self.chargers)`?

steinmn: I'm really not sure. After having set the value to 4 or 5 by accident once and needing to unplug and replug my car to get it out of whatever error mode the charging was in, I've never touched available_current directly while the car is plugged in, only through automations.

sveinse: In my charger the `charger_min_current` is 0, so I have no direct indication that the minimum is 6A. When both `charger_min_current` and `charger_max_current` is 0, which is the case when the charging is paused due to no available current in the house, then the field `allocated_charge_current` is 6A. So _maybe_ this gives a hint.

TL;DR: I'm not sure how to reliably use this to set a lower current limit.

steinmn: Ok, thanks for investigating, I guess it's better to just leave it at that.

sveinse: I have learned that PRO can have max currents up to 63A, so hardcoding 32 A is not a good idea. I did as you proposed and use `max_current` as basis for the max value in the function. Thanks.

custom_components/zaptec/api.py
```diff
@@ -546,13 +546,27 @@ async def set_limit_current(self, **kwargs):
...
             raise ValueError(
                 "Either availableCurrent or all of availableCurrentPhase1, "
                 "availableCurrentPhase2, availableCurrentPhase3 must be set"
             )
-
+        # Use 32 as default if missing or invalid value.
+        try:
+            max_current = float(self.get("max_current", 32.0))
+        except (TypeError, ValueError):
+            max_current = 32.0
```
steinmn, line 558: I'm counting 3 different ways this can default to 32, should we maybe include a `_LOGGER.debug` in the except case, or is there nothing to gain from that info?

steinmn (commented):
Just a small question, otherwise LGTM

custom_components/zaptec/api.py
sveinse: Neeh, not really. It's the case if the value is not present or is null/None, where we don't have a value to go by. I don't think its terribly important to inform, as this is only used for range checking.

Issue #270, opened 2025-08-05
update to 0.8 error


dsmagghe:
Logger: homeassistant.helpers.event
Source: helpers/template.py:643
First occurred: 13:19:39 (2 occurrences)
Last logged: 13:19:39

Error while processing template: Template<template=({% if state_attr('sensor.zaptec_charger_zpr110516', 'communication_signal_strength') | float >= -67 %} mdi:wifi-strength-4 {% elif state_attr('sensor.zaptec_charger_zpr110516', 'communication_signal_strength') | float >= -72 %} mdi:wifi-strength-3 {% elif state_attr('sensor.zaptec_charger_zpr110516', 'communication_signal_strength') | float >= -80 %} mdi:wifi-strength-2 {% elif state_attr('sensor.zaptec_charger_zpr110516', 'communication_signal_strength') | float >= -90 %} mdi:wifi-strength-1 {% else %} mdi:wifi-strength-outline {% endif %}) renders=3>
Error while processing template: Template<template=({% if state_attr('sensor.zaptec_charger_zpr223785', 'communication_signal_strength') | float >= -67 %} mdi:wifi-strength-4 {% elif state_attr('sensor.zaptec_charger_zpr223785', 'communication_signal_strength') | float >= -72 %} mdi:wifi-strength-3 {% elif state_attr('sensor.zaptec_charger_zpr223785', 'communication_signal_strength') | float >= -80 %} mdi:wifi-strength-2 {% elif state_attr('sensor.zaptec_charger_zpr223785', 'communication_signal_strength') | float >= -90 %} mdi:wifi-strength-1 {% else %} mdi:wifi-strength-outline {% endif %}) renders=3>
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/template.py", line 2446, in forgiving_float_filter
    return float(value)
TypeError: float() argument must be a string or a real number, not 'NoneType'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/template.py", line 641, in async_render
    render_result = _render_with_context(self.template, compiled, **kwargs)
  File "/usr/src/homeassistant/homeassistant/helpers/template.py", line 2977, in _render_with_context
    return template.render(**kwargs)
           ~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/jinja2/environment.py", line 1295, in render
[...]

PR #271, opened 2025-08-05
Fix redactor unable to handle sets

steinmn:
Should fix the main exception of #270, but perhaps not the underlying issue


steinmn:
I can't be 100% certain, but the main issue is that build fails due to redact not handling sets, which (I think) would lead to all entities becoming unavailable, ie value=None. All of the nonetype errors are from a template helper, so It's just triggered by the charger object not having the communication_signal_strength-attribute, which would make sense since it never got to the polling stage since build failed.

Issue #280, opened 2025-08-12
Fix nearly invisible Zaptec-logo as firmware-icon


steinmn:
I think this entails adding an entity_picture-function to the ZaptecUpdate-entity.
It currently uses the default for `UpdateEntity`:
```
    @property
    def entity_picture(self) -> str | None:
        """Return the entity picture to use in the frontend.

        Update entities return the brand icon based on the integration
        domain by default.
        """
        return (
            f"https://brands.home-assistant.io/_/{self.platform.platform_name}/icon.png"
        )
```

This gives the regular light-mode icon on the `firmware_update`-entity which is close to invisible in dark-mode. 

(Not sure what to replace it with since the backend doesn't know or care if the frontend(s) is in light or dark mode, but it seemed like a step forward to have this as an issue rather than as a `# FIXME: Find how icons work for firmware`-comment in the code. Maybe a separate fw-icon?)

sveinse:
I did some research about this some time back, and it turns out that the entity icon doesn't use the front-end UI dark mode selection. They have some system in the front-end for the main brand assets select the light and dark artwork based on the URL, but not for this type of icons. Unfortunately I cant seem to find the issue and the discussion.

sveinse:
I'm not sure we have have the options of a 3rd icon. These assets are not included into this integration. Instead they are automatically added to the HA brands repo. These are the Zaptec files https://github.com/home-assistant/brands/tree/master/custom_integrations/zaptec which is copied from the original Zaptec artwork for light and dark.

We can get creative, but the root cause is really a limitation in HA and it should be addressed there. -- I am curious thou how the other integrations handle this. Do they modify the vendor's original artwork to make it light and dark-safe?

PR #281, opened 2025-08-12
Reclassify entities from diagnostic to config where appropriate

### LINKED ISSUE #279: Ensure appropriate entity_categories
(CodeQuality) Ref [integration-quality-scale/rules/entity-category](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/entity-category/) and the associated [documentation](https://developers.home-assistant.io/docs/core/entity#registry-properties):

> Set to EntityCategory.CONFIG for an entity that allows changing the configuration of a device, for example, a switch entity, making it possible to turn the background illumination of a switch on and off. Set to EntityCategory.DIAGNOSTIC for an entity exposing some configuration parameter or diagnostics of a device but does not allow changing it, for example, a sensor showing RSSI or MAC address.

According to these definitions, I would say both `permanent_cable_lock` and `status_indicator_brightness` should be config and not diagnostic. I'm less sure about `restart_charger` and `upgrade_firmware`, but since they are controllable, I'm inclined to think they should also be config.

steinmn:
According to [integration-quality-scale/rules/entity-category](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/entity-category/) and the associated [documentation](https://developers.home-assistant.io/docs/core/entity#registry-properties)
Fixes #279

### DIFF
--- a/custom_components/zaptec/button.py
+++ b/custom_components/zaptec/button.py
@@ -81,14 +81,14 @@ CHARGER_ENTITIES: list[EntityDescription] = [
     ZapButtonEntityDescription(
         key="restart_charger",
         translation_key="restart_charger",
-        entity_category=const.EntityCategory.DIAGNOSTIC,
+        entity_category=const.EntityCategory.CONFIG,
         icon="mdi:restart",
         cls=ZaptecButton,
     ),
     ZapButtonEntityDescription(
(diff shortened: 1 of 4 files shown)


steinmn:
#280 replaces FIXME-comment

Issue #278, opened 2025-08-10
resume_charging in new release


gustafssone:
Using Version 0.8.1. In the release notes resume_charging is said to be replaced by a button. It used to already be a button (in my case), however it is most of the time unavailable. 
Do I need to remove it in order to recreate it?

"button.garage01_stop_charging" is intermittently available (when cable is connected, otherwise unavailable), but button.garage01_resume_charging goes from Unknown to Unavailable. 

switch.garage01_charging, same behavior as "stop".

sveinse:
The reason these buttons become unavailable is because Zaptec has put restrictions on us to not fire off the commands if we're not in the right mode. Thus stop charging should only be available when charging the car. Unfortunately, HA doesn't have "disabled" buttons, so making them unavailable is the only means we have.

However, I'm seeing the same thing: "Resume charging" button is not available when my car is in mode "Charge done", which I think is a bug.

@gustafssone when you were testing, what mode is the charger in when the button is available/unavailable? The "Charger mode" entity indicates the mode.

steinmn:
`resume_charging` is only available if both the mode is "Charge done" and `FinalStopActive` is 1. While I'm not 100% sure about the behaviour of `FinalStopActive`, I expect it means whether or not the `stop_charging` command is responsible for the charging being stopped. If charging was not stopped with the `stop_charging` command, `resume_charging` will not work.

Some examples where mode will be "Charge done" and `resume_charging` will not be able to start the charging:
- The charging has stopped because the car has reached its configured target battery level
- The charging has stopped because it was stopped using the app/api of the car
- The charging has stopped because the time is outside of the charging schedule configured in the app of the car

While setting the buttons as "unavailable" is not ideal, it is the only way (we found) of giving a visual feedback to the user that the action will not be permitted. If we did not do this, pressing the button would just trigger an error message from the Zaptec API rejecting the command.

gustafssone:
So in my case I just activated Grid Rewards from Tibber. They will start and stop the charging a few times during the night. I can of course revoke their permissions, but some users may have external triggers for charging. Not sure on how to deal with it.

steinmn:
> Do the user expect to be able to resume a charge when stopped by an external source?

If it's their first time charging a car on a smart-charger, yeah probably. At least I think that was what I thought the first time I tried charging my car. In any case, I think the length of this discussion shows we should add something about it in the readme.

[...]

Issue #192, opened 2025-07-14
Missing API features from Zaptec cloud


sveinse:
The [Zaptec API fair use policy](https://docs.zaptec.com/docs/api-fair-use-policy#/) forbids integrators to use features that are not part of the official API. We have been using a couple of those in this integration, which we probably will have to remove. These are functions which users ask for.

The purpose of this issue is to collect a list of functionality we in the Zaptec Home Assistent community would like to request Zaptec to add to their official API. I can't promise they will listen, but we can attempt since the HA user base is quite large.

I'll update this description as items are added in the discussions.

1. **Push notifications**
    * No-one likes polling. Doing it too frequent puts strains on the cloud, doing it too infrequent gives delays in data updates in HA.
    * We wish for a non-deprecated mechanism to subscribe to push notifications on data updates to give near-realtime updates of the data.
    * Fast updates is especially important in HA where data changes might further be used as triggers in user automations

2. **Delayed start**
    * Prevent automatic charge start when connecting a car to the charger
    * Can be handled by setting max current to 0, but these are observed to reset when not charging. Zaptec does not recommended to change these values when not connected to car.
    * Use-cases described in [#49](https://github.com/custom-components/zaptec/issues/49#issuecomment-1668673302)
    
3. **Authorize from integration**
    * Support the trigger of authorization from HA
    * Used by some as a means to prevent auto start of charging and keeping the cable locked until authorized
    * Discussed in #55 

4. **Permanent cable lock**
    * Some users want to automate when the cable locking from HA. Today this has to be done via the Portal.
    * Issue #102 

5. **HMI brightness**
    * The ability to set the HMI brightness without needing to use the Portal or app
    * Issue #112

Lheriss:
Hi,
I would like to have a way to identify which rfid tag is used to start a session. I have configured three of them, and would like to be able to detect how much energy is used with each of them...
Thanks for your incredible work!

sveinse:
Added item 7 enable/disable ECO mode with reference to #153.

sveinse:
@thecoldwine This issue contains the list of features the HA community would like to make to Zaptec. They are based on actual use and feature requests made by our users. I hope you can consider these, please. As mentioned, our install base is ~2k-4k installations (of HA).

sveinse:
Thanks @steinmn I updated the top description to reflect your input.

thecoldwine:
[...]

PR #282, opened 2025-08-13
Add explanation of resume_charging-limitations

steinmn:
The stop/resume commands have some limitations that aren't very intuitive if you haven't struggled directly with the API (and maybe not even then). Added some clarification to the readme to hopefully reduce questions/confusion and avoid issues like #278.

Also correct some misnamed deprecated services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,21 +170,46 @@ Starting and stopping charging can be done by several methods. If the charger
 is configured to not require authentication, connecting the charger to the
 EV will by default start charging.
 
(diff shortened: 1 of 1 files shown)


sveinse:
Jep. But I'd like to check if they are doing it differently. If they are, we'd like to do so as well and add it to our feedback.

steinmn:
> In the Zaptec App it sais "Charging stopped by car", which is not information this integration has.

Goes on the #192 list?

PR #284, opened 2025-08-15
Avoid limit_current service call passing None-values and clarify error messages

### LINKED ISSUE #283: Error when calling limit_current: '<=' not supported between instances of 'int' and 'NoneType' after recent update
After updating to the latest version of the Zaptec integration, the service call zaptec.limit_current fails both in Node-RED and directly from Home Assistant Developer Tools.

Actual behavior:
The service fails with the following error: HomeAssistantError: Limit current failed: '<=' not supported between instances of 'int' and 'NoneType'

Additional information:
This setup has worked reliably for over two years.
The error occurs even when calling the service directly from Developer Tools, without Node-RED.
In Developer Tools → States, the charger entity no longer shows a numeric max_current attribute — it appears to be null.
It seems this attribute might have been removed in the latest version, but the limit_current logic still expects it to be a number.

Home Assistant core version: 2025.8.1

Zaptec integration version: 0.8.1

steinmn:
Fixes #283

### DIFF
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -551,14 +551,13 @@ class Installation(ZaptecBase):
         ]
         if not (has_availablecurrent ^ all(has_availablecurrentphases)):
             raise ValueError(
-                "Either availableCurrent OR all of availableCurrentPhase1, "
+                "Either availableCurrent or all of availableCurrentPhase1, "
                 "availableCurrentPhase2, availableCurrentPhase3 must be set"
             )
-        if has_availablecurrent and any(has_availablecurrentphases):
+        if any(has_availablecurrentphases) and not all(has_availablecurrentphases):
             raise ValueError(
-                "A combination of both availableCurrent and any of "
-                "availableCurrentPhase1, availableCurrentPhase2, "
-                "availableCurrentPhase3 is not allowed, choose one"
+                "If any of availableCurrentPhase1, availableCurrentPhase2 and "
+                "availableCurrentPhase3 are set, then all of them must be set"
             )
 
         # Use 32 as default if missing or invalid value.
--- a/custom_components/zaptec/api.py
+++ b/custom_components/zaptec/api.py
@@ -541,19 +541,26 @@ class Installation(ZaptecBase):
         availableCurrentPhase* to set each phase individually.
         """
         has_availablecurrent = kwargs.get("availableCurrent") is not None
-        has_availablecurrentphases = all(
+        has_availablecurrentphases = [
             kwargs.get(k) is not None
             for k in (
                 "availableCurrentPhase1",
                 "availableCurrentPhase2",
                 "availableCurrentPhase3",
             )
-        )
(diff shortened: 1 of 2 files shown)


custom_components/zaptec/services.py
```diff
@@ -279,19 +279,20 @@ async def service_handle_upgrade_firmware(service_call: ServiceCall) -> None:
...
+        limit_args = {}
+        # only add the relevant arguments if they are not None
+        if available_current := service_call.data.get("available_current"):
+            limit_args["availableCurrent"] = available_current
+        if available_current_phase1 := service_call.data.get("available_current_phase1"):
+            limit_args["availableCurrentPhase1"] = available_current_phase1
+        if available_current_phase2 := service_call.data.get("available_current_phase2"):
+            limit_args["availableCurrentPhase2"] = available_current_phase2
+        if available_current_phase3 := service_call.data.get("available_current_phase3"):
+            limit_args["availableCurrentPhase3"] = available_current_phase3
```
sveinse: This won't work if the value is being set to 0. It needs to be explicitly tested for `None`.

custom_components/zaptec/api.py
```diff
@@ -541,19 +541,26 @@ async def set_limit_current(self, **kwargs):
...
                 "availableCurrentPhase2",
                 "availableCurrentPhase3",
             )
-        )
-        if not (has_availablecurrent ^ has_availablecurrentphases):
+        ]
+        if not (has_availablecurrent ^ all(has_availablecurrentphases)):
             raise ValueError(
-                "Either availableCurrent or all of availableCurrentPhase1, "
+                "Either availableCurrent OR all of availableCurrentPhase1, "
```
sveinse: I'm not overly fond of using capital letters in the middle of an English sentence

custom_components/zaptec/api.py
```diff
@@ -541,19 +541,26 @@ async def set_limit_current(self, **kwargs):
...
                 "availableCurrentPhase2, availableCurrentPhase3 must be set"
             )
+        if has_availablecurrent and any(has_availablecurrentphases):
+            raise ValueError(
+                "A combination of both availableCurrent and any of "
+                "availableCurrentPhase1, availableCurrentPhase2, "
+                "availableCurrentPhase3 is not allowed, choose one"
+            )
```
sveinse: This error is a little confusing to me. Perhaps it's better to test that if any of the three (with `any()`) have used, and all of them must be present. Error: "All of availableCurrent1, availableCurrent2 and availableCurrent2 must be present".

custom_components/zaptec/services.py
steinmn: :facepalm: tried to be a little bit too elegant, fixed now

custom_components/zaptec/api.py
steinmn: Wanted to add emphasis to make it clearer that it's an exclusive or, but I'm fine either way

custom_components/zaptec/api.py
steinmn: Modified

PR #288, opened 2025-08-17
Add breaking changes warning about blueprint

steinmn:


README.md
```diff
@@ -412,6 +413,12 @@ updated. The English mapping of the changed values is
 - disconnected:         Disconnected
 - unknown:              Unknown
 
+The changes to the Charger mode will also cause the
+[Load Balancing blueprint](#load-balancing-your-charger) compatible with 0.7.x
+to stop working. There is an [open PR](https://github.com/svenakela/ha/pull/10)
+for a 0.8.x-compatible version. If you use this blueprint, or a variant of it,
+you will need to update your blueprint/automation accordingly.
```
sveinse, line 420: Perhaps add an information box that the blueprint is language dependent, and that it might need modification if not using English?

sveinse (commented):
This looks good. Added one small comment.

README.md
steinmn: I think that belongs in the
[blueprint readme](https://github.com/svenakela/ha/blob/main/zaptec/README.md), not in this repo. We already refer to that for "How to set up the automation, how the logic works and what all settings mean".

sveinse: Yes, I agree. However, it seems we're getting the issues about it, right?

steinmn: Well in this case we were the ones who broke it, so that's fair imo.

On the other hand, I see we have a blueprint issue in https://github.com/custom-components/zaptec/issues/179, but again, I think that belongs in the svenakela/ha-repo, not here. If there's anything that needs to be done inside the integration to fix it, then sure, but the issue with #179 is purely from the setup of the blueprint.

PR #298, opened 2025-08-22
Disable ruff errors that HA core is also ignoring

steinmn:
Simple example from HA core that violates both of the boolean rules: https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/entity_platform.py#L79

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -28,6 +28,10 @@ ignore = [
 
     "TRY003", # Avoid specifying long messages outside the exception class
     "TRY400", # Use `logging.exception` instead of `logging.error`
+
+    # definitely skipped in HA, not obvious how
+    "EM101",  # Exception must not use a string literal, assign to variable first
+    "EM102",  # Exception must not use an f-string literal, assign to variable first
 ]
 
 [lint.flake8-pytest-style]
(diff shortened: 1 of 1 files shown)


.ruff.toml
```diff
@@ -28,6 +28,12 @@ ignore = [
 
     "TRY003", # Avoid specifying long messages outside the exception class
     "TRY400", # Use `logging.exception` instead of `logging.error`
+
+    # definitely skipped in HA, not obvious how
+    "EM101",  # Exception must not use a string literal, assign to variable first
+    "EM102",  # Exception must not use an f-string literal, assign to variable first
+    "FBT001", # Boolean-typed positional argument in function definition
+    "FBT002", # Boolean default positional argument in function definition
```
sveinse, line 36: HA is not only doing exclusion of rules, but also inclusion (selection). This means that there might be rules that aren't selected https://github.com/sveinse/ha-core/blob/facf217b995982ff5c504bd59d419a59c31ee068/pyproject.toml#L645

steinmn: It also looks like there's some separate rules for pylint as well. Not sure how everything is set up on the HA side, but I'm thinking we focus on fixing/excluding the errors we currently have until we can turn off `--exit-zero`, and then we can start considering adding more rules.

PR #307, opened 2025-08-28
Add validation and monitoring of requirement compatibility with HA

### LINKED ISSUE #305: Set up tests/workflows to catch dependency issues early
Cases we want to cover:

1. Home Assistant Core increases the required version of a package so that it violates the requirements in our manifest.json.
2. We increase the required version of a package in our manifest.json so that we need to also increase the minimum version of Home Assistant Core

steinmn:
Resolves #305 

There are 2 checks:

1. Checking if the current manifest.json requirements are compatible with the latest dev branch of HA.
2. Checking if the current manifest.json requirements are compatible with all HA versions newer than the minimum HA-version listed in hacs.json

1 is run on every push and PR in validate.yaml, and also periodically in monitor-requirements.yaml.
2 is only run on push and PR if there are changes made to either manifest.json and/or hacs.json

Example runs:

- [No changes to manifest/hacs](https://github.com/steinmn/zaptec/actions/runs/17290169435/job/49075446675)
- [Acceptable changes to manifest/hacs](https://github.com/steinmn/zaptec/actions/runs/17290253297/job/49075705465)
- [Unacceptable changes to manifest/hacs](https://github.com/steinmn/zaptec/actions/runs/17290314175/job/49075900528)

### DIFF
--- a/.github/workflows/monitor-requirements.yaml
+++ b/.github/workflows/monitor-requirements.yaml
@@ -17,12 +17,3 @@ jobs:
         run: |
           cd ${{ github.workspace }}
           python3 scripts/check_requirements.py --current-ha-dev
-
-      # Consider adding auto-generated issues, need to make sure we don't create duplicates first
-      # - name: Create issue if incompatible
-      #   if: failure()  # only run if previous step exited 1
-      #   uses: peter-evans/create-issue-from-file@v5
-      #   with:
-      #     title: "Requirements incompatibility detected"
-      #     content-filepath: ./issue.md
-      #     labels: bug, compatibility


.github/workflows/monitor-requirements.yaml
```diff
@@ -0,0 +1,28 @@
...
+          ./check_requirements.py --current-ha-dev
+
+      # Consider adding auto-generated issues, need to make sure we don't create duplicates first
+      # - name: Create issue if incompatible
+      #   if: failure()  # only run if previous step exited 1
+      #   uses: peter-evans/create-issue-from-file@v5
+      #   with:
+      #     title: "Requirements incompatibility detected"
+      #     content-filepath: ./issue.md
+      #     labels: bug, compatibility
```
sveinse: What's the intention of leaving this commented code in the file?

.github/workflows/validate.yaml
```diff
@@ -44,4 +44,30 @@ jobs:
         uses: astral-sh/ruff-action@v3
         with:
           args: check --exit-zero # succeed despite errors until https://github.com/custom-components/zaptec/issues/258 is done
-          src: "."
+          src: "./custom_components/zaptec"
+
+  manifest-requirements:
```
sveinse, line 53: Since we're extended to more tests, should we also reduce when this script is run? I believe this workflow will run even on push to forks. It's a common choice to do it on PRs and on merge to main/master, I think.

.github/workflows/validate.yaml
```diff
@@ -44,4 +44,30 @@ jobs:
...
+    name: Manifest requirements validation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout zaptec repo
+        uses: actions/checkout@v4
+
+      - name: Check manifest requirements against latest HA dev branch
+        run: |
+          cd ${{ github.workspace }}/scripts
+          ./check_requirements.py --current-ha-dev
```
sveinse: Wouldn't it be more conventional to expect the root of the repo as the PWD and not scripts? I see in the py code that `parents[]` is used to get the root.

Running python script via shebangs isn't always stable and portable. E.g. if we are at some point going to run on other platforms.

E.g.
```sh
cd ${{ github.workspace }}
python3 scripts/check_requirements.py --current-ha-dev
```
I think this looks neater

scripts/check_requirements.py
```diff
@@ -0,0 +1,192 @@
...
+ROOT = Path(__file__).resolve().parents[1]
+MANIFEST = ROOT / "custom_components" / "zaptec" / "manifest.json"
+HACS = ROOT / "hacs.json"
+CONSTRAINTS_URL = "https://raw.githubusercontent.com/home-assistant/core/dev/homeassistant/package_constraints.txt"
+CONSTRAINTS_V_URL = "https://raw.githubusercontent.com/home-assistant/core/{version}/homeassistant/package_constraints.txt"
+HA_RELEASES_URL = "https://api.github.com/repos/home-assistant/core/tags"
+PYPI_URL = "https://pypi.org/pypi/{package}/json"
+
+
+def load_manifest_requirements() -> list[Requirement]:  # noqa: D103
```
sveinse: I got hung up with the `noqa: D103` statement that's repeated in every function. I had to google to find what it was. Perhaps this deserves a comment what D103 is so its plain to see?

The elephant in the room here is: Would it be better to actually write those docstrings?

scripts/lint
```diff
@@ -7,10 +7,10 @@ cd "$(dirname "$0")/.."
 if [ "$1" = "ci" ]; then
     # Don't try to fix errors, just fail
     ruff format . --diff
-    ruff check . --exit-zero # succeed despite errors until https://github.com/custom-components/zaptec/issues/258 is done
+    ruff check ./custom_components/zaptec --exit-zero # succeed despite errors until https://github.com/custom-components/zaptec/issues/258 is done
```
sveinse: The `./` is superflous. `ruff check custom_components/zaptec` should do it.

sveinse (commented):
Some comments and nitpicks. Generally it looks good. I haven't done any verification of the functionality of the python file, just been looking at the code structure.

.github/workflows/monitor-requirements.yaml
steinmn: Original intention was to auto-generate an issue when detecting incompatibilitties, but I chickened out when I realised the potential for duplicates. Do you get email-notifications on failed actions for the main zaptec repo? Do you think we should spin this out as a separate issue or just discard the idea?

scripts/check_requirements.py
steinmn: Eh :shrug: Are there any of the functions that aren't fully explained by the name/inputs? Since I ended up disabling ruff check for this file, I'm tempted to just remove the noqa-s and let ruff complain in the file.

.github/workflows/validate.yaml
steinmn: Are there downsides to running this? Unless there are changes to manifest.json or hacs.json (triggering the second check), this is the fastest of the tests. Also, more runs means new incompatibilities from the HA side are caught earlier.

.github/workflows/validate.yaml
steinmn: Done

scripts/check_requirements.py
sveinse: What I meant is that I had to google what `D103` is to understand what it excepts. It would be nice with a simple comment saying that D103 will disable docstrings requirements.

steinmn: Added a comment on the first D103.
Sidenote: In VSCode you get the full ruff explanation if you just hover the mouse over D103
image

sveinse: Yeah, as I'm going over the ruff errors in other places of the code, I see that we sometimes need to except an error and that it probably needs to be expected that one is able to look them up.

.github/workflows/monitor-requirements.yaml
sveinse: No, I'm not owner in the "custom-components" organization unfortunately, so I don't get these emails.

.github/workflows/validate.yaml
sveinse: It will run on all commits, even in users forks.

I have noticed that I often get a ❌ on my own commits when I do PRs, but before the actions have run in this repo. After a minute, the commits succeed ✅. Why is that?

I found out that the PR is running two sets of GHA jobs. One in my own personal repo where I stage the PR from, and one in this custom_components/zaptec repo. The first fails, the second succeeds:

* My personal: https://github.com/sveinse/zaptec/actions/runs/17358803514
* In this repo: https://github.com/custom-components/zaptec/actions/runs/17358873214

I'm not sure you're permitted to view these. I do get failed emails for the job running in my repo

So the conclusion is that we should probably adjust the `on` triggers in the workflow file to run when and where it should. For reference the HACS integration blueprint repo is suggesting this: https://github.com/ludeeus/integration_blueprint/blob/51ec561486e60cbf565dda4161ddab12aeb424bc/.github/workflows/validate.yml#L3-L12

.github/workflows/monitor-requirements.yaml
sveinse: I might not be correct: I'm getting emails from custom_component/zaptec

.github/workflows/validate.yaml
steinmn: > I have noticed that I often get a ❌ on my own commits when I do PRs, but before the actions have run in this repo. After a minute, the commits succeed ✅. Why is that?

I'm guessing the push trigger on your repo (which fail) is counted first, and then once the PR-trigger in this repo succeeds, that overwrites the status? also, looks like you are missing topics in your fork like I did before, not sure if it's because topics aren't included in forks or if we just forked before the topics were added.

Added the restriction on push+PR. If I understand the schedule-trigger correctly, that only runs on the master branch anyway.

.github/workflows/monitor-requirements.yaml
steinmn: Separated out into #315 and removed comment

sveinse: Yeah, I agree thats probably best for now. Thanks.

.github/workflows/validate.yaml
sveinse: Great. Let's try it as proposed now.

PS! The topics have been there for ages, so it doesn't pass along when forking. At least not when we did so.

sveinse (approved):
I think this looks good now. Approved.

Unless there is something you want added here last minute?

PR #313, opened 2025-08-31
Adding pytest to project

### LINKED ISSUE #262: Add pytest setup

sveinse:
Fix #262

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -7,9 +7,6 @@ on:
 
 permissions: {}
 
-env:
-  DEFAULT_PYTHON: 3.13
-
 jobs:
   hassfest: # https://developers.home-assistant.io/blog/2020/04/16/hassfest
     name: Hassfest validation
@@ -59,7 +56,7 @@ jobs:
       - name: Setup Python {{ "${{ env.DEFAULT_PYTHON }}" }}
         uses: "actions/setup-python@v5"
         with:
-          python-version: {{ "${{ env.DEFAULT_PYTHON }}" }}
+          python-version: '3.13'
 
       - name: Install requirements
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/validate.yaml
```diff
@@ -45,3 +45,25 @@ jobs:
         with:
           args: check --exit-zero # succeed despite errors until https://github.com/custom-components/zaptec/issues/258 is done
           src: "."
```
steinmn: Do we want to include the tests in the ruff check? (ref modification to exclude scripts in #307)

sveinse: (I was about to ask you the same question.) That up to us, really. I've been working with projects that have both with or without ruff in tests/. I think the argument goes something along the lines that the test code is an equal important piece of the quality of the code, hence it should have the same scrutiny as the main code. The other argument goes that tests often require some non-pythonic constructs and that following every rule in tests is being somewhat pedantic.

I'm leaning towards not including tests into the ruff check (for now at least). But I'm very keen on hearing your opinion before we settle this.

steinmn: From what I can tell, HA-core runs it on all files. So I'm inclined to say include tests for now, and if it gets really annoying/tedious, we can consider excluding it again.

sveinse: Yes. Smart to check what HA does. Then I agree, let's leave it to also check tests.

sveinse:
> LGTM, will probably need a little bit of coordination to add both this and #307 to validate.yaml

I plan to to the merges in order, so #307 goes first and then I'll fixup any merge conflicts that might result in this PR.

Are there any data content we need to coordinate as well here?

sveinse:
I added ruff check of tests to this PR. Please take a look at the results from the latest run. I see a few errors from ruff that we have to address to not get swamped by warnings. E.g. that assert has been encountered (which is expected in pytest).

PR #314, opened 2025-08-31
Move ZCONST into const to make it importable

sveinse:
* Move ZCONST initialization from api to const
* Use ZCONST in Redactor instead of class argument

By moving ZCONST initialization from api.py to const.py, it will be easier to use ZCONST from anywhere in the code. One example is what's included in the PR: The redact.py can now use ZCONST directly and doesn't need to get the data passed on via the class instance initialization.

### DIFF
--- a/custom_components/zaptec/diagnostics.py
+++ b/custom_components/zaptec/diagnostics.py
@@ -56,7 +56,7 @@ async def _get_diagnostics(
     zaptec: Zaptec = manager.zaptec
 
     # Helper to redact the output data
-    redact = Redactor(DO_REDACT, ZCONST.observations)
+    redact = Redactor(DO_REDACT)
 
     def add_failure(err: Exception) -> None:
         """Add a failure to the output."""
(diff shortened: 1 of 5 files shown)


custom_components/zaptec/zaptec/zconst.py
```diff
@@ -194,3 +194,7 @@ def type_user_roles(self, val: int) -> str:
         # v != 0 is needed to avoid the 0 == 0 case leading to None always being included
         roles = {k for k, v in self.get("UserRoles", {}).items() if v != 0 and (v & val) == v}
         return ", ".join(roles)
+
+
+ZCONST = ZConst()
+"""Zaptec constants instance."""
```
steinmn: While this might be more accurate than the old "Global var for the API constants from Zaptec"-explanation, the old one at least gave me a clearer impression of what it was for and where it came from.

I think it would be worth it to expand the docstring a bit here, considering this is also used outside of the API-folder and it's nice to get that explanation on a mousover:
image

sveinse: Jep. I agree. docstring updated.

steinmn (approved):

PR #316, opened 2025-08-31
Readd ruff check for scripts, but allow print and missing docstring

steinmn:


.ruff.toml
```diff
@@ -8,6 +8,10 @@ line-length = 98
     "S101",  # Allow assert in tests
     "INP001",  # Don't complain over missing __init__.py in test dirs
 ]
+"scripts/*" = [
+    "D103",  # Don't complain over missing docstrings in scripts
+    "T201",  # Allow for scripts to write to stdout
+]
```
sveinse: I regret placing the `[lint.per-file-ignores]` above the general `[lint]` section. Would you be so kind to move this entire section to below ´[lint]` please? I think it's best to have the main rules first in the files.

steinmn: I ended up reordering the sections to match the order in https://github.com/home-assistant/core/blob/dev/pyproject.toml.

I also added the pydocstyle section from HA-pyproject, which apparently disables the `D401 First line of docstring should be in imperative mood`-ruff-error (which I think we have discussed ignoring before at some point)

sveinse (approved):
I think this looks good. Are we ready to merge?

steinmn:
Yep :+1:

PR #317, opened 2025-08-31
Cleanup misc - add typing and tests

sveinse:
This PR contains the cleanup of zaptec/misc.py. All ruff errors have been removed (and even some caught by the pyright static type checker) and tests for the functions have been added.


steinmn:
Could we maybe rename the file to utils.py or something instead? Misc just seems like a draft to me.

PR #318, opened 2025-09-02
Add unit tests for zconst and print test coverage in gh-action

steinmn:
(bonus: tests caught a typo in zconst.py)

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+# run tests with -s to display printouts
+pytest -s --cov=./custom_components/zaptec
+
+# generate coverage report in html
+coverage html
--- a/tests/zaptec/test_zconst.py
+++ b/tests/zaptec/test_zconst.py
@@ -1,0 +1,180 @@
+"""Tests for zaptec/zconst.py."""
+
+import asyncio
+
+import pytest
+
+from custom_components.zaptec.zaptec.api import Zaptec
+from custom_components.zaptec.zaptec.zconst import ZCONST
+
+
+@pytest.fixture
+def zaptec_constants() -> dict:
+    """Get latest constants from Zaptec API."""
+
+    async def get_zaptec_constants() -> dict:
+        async with Zaptec("username", "password") as zaptec:
+            # the constants API endpoint does not require login
+            const: dict = await zaptec.request("constants")
+            return const
+
+    return asyncio.run(get_zaptec_constants())
+
+
+@pytest.mark.dependency
+def test_init_zconst(zaptec_constants: dict) -> None:
+    """Test the initialization of the ZCONST instance."""
+    ZCONST.clear()
+    ZCONST.update(zaptec_constants)
+    ZCONST.update_ids_from_schema(None)
+
+
+@pytest.mark.dependency(depends=["test_init_zconst"])
+def test_charger_operation_modes() -> None:
+    """Test the contents of charger_operation_modes_list."""
+    assert "Unknown" in ZCONST.charger_operation_modes_list
+    assert "Disconnected" in ZCONST.charger_operation_modes_list
+    assert "Connected_Requesting" in ZCONST.charger_operation_modes_list
+    assert "Connected_Charging" in ZCONST.charger_operation_modes_list
+    assert "Connected_Finished" in ZCONST.charger_operation_modes_list
+    assert len(ZCONST.charger_operation_modes_list) == 5  # noqa: PLR2004
+
+    assert ZCONST.type_charger_operation_mode(1) == "Disconnected"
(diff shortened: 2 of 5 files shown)


scripts/test
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+# run tests with -s to display printouts
+pytest -s --cov=./custom_components/zaptec
+
+# generate coverage report in html
+coverage html
```
sveinse: How do you view these reports? I've had problems getting a usable browser within the dev container.

...so I've reverted to using xml and a coverage view tool in VS Code, "Coverage gutters". Notably, this extension _is_ installed in our dev container.

I guess the question is: Do we want html here or some other, like xml - or both?

tests/zaptec/test_zconst.py
```diff
@@ -0,0 +1,180 @@
...
+from custom_components.zaptec.zaptec.zconst import ZCONST
+
+
+@pytest.fixture
+def zaptec_constants() -> dict:
+    """Get latest constants from Zaptec API."""
+
+    async def get_zaptec_constants() -> dict:
+        async with Zaptec("username", "password") as zaptec:
+            # the constants API endpoint does not require login
```
sveinse: Perhaps better to use `async with Zaptec("N/A", "N/A") as zaptec:` to be more clear that this is not actual doing anything. I like the comment thou.

tests/zaptec/test_zconst.py
```diff
@@ -0,0 +1,180 @@
...
+def test_init_zconst(zaptec_constants: dict) -> None:
+    """Test the initialization of the ZCONST instance."""
+    ZCONST.clear()
+    ZCONST.update(zaptec_constants)
+    ZCONST.update_ids_from_schema(None)
+
+
+@pytest.mark.dependency(depends=["test_init_zconst"])
+def test_charger_operation_modes() -> None:
+    """Test the contents of charger_operation_modes_list."""
```
sveinse: I'm wondering: Would it be simpler, i.e. less typing, to make a module scoped fixture which does the same as `test_init_zconst()`? We do need `test_init_zconst()` for the test in addition of course. I personally find `@pytest.mark.dependency(depends=...)` to be tedious.

scripts/test
steinmn: I just opened index.html in my regular Chrome browser (outside of the container). 

Unless there's any incompatibilities, why not just do both and leave it to personal preference for which you want to look at?

tests/zaptec/test_zconst.py
steinmn: Ended up setting the zaptec_constants-fixture as scoped (to only call API once) and make the init function autouse to ensure its run for every test without having to include it as an input (equally as tedious as dependency, but also adding missing-annotation- and unused-rufferrors).

I think this is both simpler and better than the dependency approach, since we end up with a clean slate for each test, and we could do different tests with different `update_ids_from_schema` set inputs if we want to (haven't added that since I don't know what to test for, but the option is there at least).

scripts/test
steinmn: Added xml generation to the test script and instructions to [development.md](https://github.com/steinmn/zaptec/blob/test-zconst/DEVELOPMENT.md#running-tests)

tests/zaptec/test_zconst.py
```diff
@@ -8,28 +8,34 @@
...
             # the constants API endpoint does not require login
             const: dict = await zaptec.request("constants")
             return const
 
     return asyncio.run(get_zaptec_constants())
 
 
-@pytest.mark.dependency
+@pytest.fixture(autouse=True)  # runs for every test
+def init_zconst(zaptec_constants: dict) -> None:
```
sveinse, line 25: I like the `autouse=True` option 😊

I'm not 100% sure, but I wonder that this fixture is rerun per function test. I think this needs to be labelled with `scope="module"` too.

scripts/test
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+# run tests with -s to display printouts
+pytest -s --cov=./custom_components/zaptec
```
sveinse: While at this: Should we enable branch coverage testing as well: `--cov-branch`?

tests/zaptec/test_zconst.py
steinmn: As mentioned in the comment, it does run for every function test, but I consider that a feature and not a bug. It means ZCONST remains the same at the start of each test, regardless of what values `update_ids_from_schema` might be called with in any of the tests and in which order the tests are run. The `scope="module"` on zaptec_constants still means we only call the API once.

sveinse: Good. I now agree.

PR #324, opened 2025-09-23
Skip attribute initialisation when firmware info is missing

feliciaan:
Skip attributes and add warning for missing firmware info in charger attributes, prevents the integration from crashing while starting, if these attributes are not available.

This only happens for chargers that are added to the Zaptec portal, but not configured or correctly added.

### DIFF
--- a/custom_components/zaptec/zaptec/api.py
+++ b/custom_components/zaptec/zaptec/api.py
@@ -324,7 +324,10 @@ class Installation(ZaptecBase):
                 ):
                     # If the charger is already added to the Zaptec platform but not yet
                     # initialized, these fields are not available.
-                    _LOGGER.warning("Missing firmware info for charger %s", charger.qual_id)
+                    _LOGGER.warning(
+                        "Missing firmware info for charger %s because the charger hasn't been initialized yet. Safe to ignore.",  # noqa: E501
+                        charger.qual_id,
+                    )
                     continue
 
                 charger.set_attributes(


custom_components/zaptec/zaptec/api.py
```diff
@@ -317,6 +317,12 @@ async def poll_firmware_info(self) -> None:
                 charger = self.zaptec.get(fm["ChargerId"])
                 if charger is None:
                     continue
+                if fm.get("CurrentVersion") is None:
```
steinmn: Maybe do the check for all 3 keys just to be sure there's no obscure case where one key is defined while the others aren't?

sveinse:
@feliciaan Is this issue critical enough to delay the release of 0.8.3 ? Should we wait for this one?

custom_components/zaptec/zaptec/api.py
```diff
@@ -317,6 +317,16 @@ async def poll_firmware_info(self) -> None:
...
                 if charger is None:
                     continue
+                if (
+                    fm.get("CurrentVersion") is None
+                    or fm.get("AvailableVersion") is None
+                    or fm.get("IsUpToDate") is None
+                ):
+                    # If the charger is already added to the Zaptec platform but not yet
+                    # initialized, these fields are not available.
+                    _LOGGER.warning("Missing firmware info for charger %s", charger.qual_id)
```
sveinse: This is a warning that will be pushed to be visible to the users. Are we certain the error message is explanatory enough without the context? Can we foresee a issue report from users on this message? -- Perhaps something like "Missing firmware info from charger `%s` because the charger haven't been initialized yet. Safe to ignore" or something alike.

steinmn: @feliciaan Please respond, so that we can include this in the upcoming 0.8.4 release.

feliciaan: Changed it, I think it's ready to merge.

PR #326, opened 2025-09-24
Cleanup main readme and reduce duplication of release notes

steinmn:
With 0.8.2 having been out for over a month with little to no (self-created) issues, I think we can replace the the "not extensively tested"-warning-sign with a more generic "we are only able to test so much"-warning.

There was also a lot of duplication of the release notes that I think can be removed (but open to discuss it).

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ Confirmed to work with Zaptec products
 * Zaptec Home
 * Zaptec PRO
 
+> [!IMPORTANT]
+> The testing of this integration is limited by the developers only being able
+> to test on their own personal chargers, so your feedback is valuable.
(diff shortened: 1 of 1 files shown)


sveinse (commented):
I agree in this direction. The docs and the landing page should be a living document and I think we're ready for this move.

Unfortunately I don't have time to proof-read this today.

steinmn:
NP, hope you have time over the weekend. Judging by #329/#330, we need to get 0.8.3 out before HA 2025.10 is released.

steinmn:
According to upcoming events on https://developers.home-assistant.io/, 2025.10.0 is scheduled this wednesday (oct 1st).

PR #327, opened 2025-09-24
Move old API test code into pytest framework

steinmn:
but disable it on Github actions since it requires login credentials.


tests/zaptec/test_diagnostics.py
```diff
@@ -0,0 +1,79 @@
...
+
+from custom_components.zaptec.diagnostics import _get_diagnostics
+from custom_components.zaptec.zaptec.api import Zaptec
+
+# Always set to "true" when GitHub Actions is running a workflow
+IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
+
+# Set manually, or when running 'scripts/test --skip-api'
+SKIP_API_TEST = os.getenv("SKIP_ZAPTEC_API_TEST") == "true"
+
```
sveinse: It would be better to move them to conftest.py and make a proper mark. This is probably useful to several test function.

tests/zaptec/test_diagnostics.py
```diff
@@ -0,0 +1,79 @@
...
+            config = FakeConfig(
+                runtime_data=manager,
+            )
+            hass = None
+
+            # Get the diagnostics info
+            out = await _get_diagnostics(hass, config)
+            _LOGGER.info(out)
+
+    asyncio.run(gogo())
```
sveinse: I think it be better to use pytest-asyncio and make it a `async def test_*` function. Using asyncio.run() in pytest is not ideal.

steinmn: Done :+1:

tests/zaptec/test_diagnostics.py
steinmn: Could you please clarify what you mean by "proper mark"? I'm probably too tired right now, but I just couldn't get fixtures and mark to behave like I wanted them to. Will take another look at it over the weekend.

steinmn: A good nights sleep and some sparring with ChatGPT worked wonders, please have a look at the updated version

sveinse:
This is how `mark` can be customized:

```py
# conftest.py
def pytest_addoption(parser):
    parser.addoption(
        "--additional", action="store_true", default=False, help="Run additional Zaptec tests"
    )

def pytest_collection_modifyitems(config, items):
    skip_additional = None
    if not config.getoption("--additional"):
        skip_additional = pytest.mark.skip(reason="Skip test when --additional isn't used")
    for item in items:
        if skip_additional is not None and "additional" in item.keywords:
            item.add_marker(skip_additional)
```

This can now be used with mark in tests:

```py
@pytest.mark.additional
def test_some_function():
    ....
```

This results in a new command line option `--additional` which runs the test functions that have been marked. Otherwise it should not. Note I haven't tested the above code, so it might not work, but this is the gist of it. https://docs.pytest.org/en/stable/how-to/writing_hook_functions.html

pyproject.toml
```diff
@@ -1,10 +1,15 @@
...
+
 [tool.pytest.ini_options]
 pythonpath = [
     ".",
 ]
 testpaths = [
     "tests",
 ]
+log_format = "%(asctime)s.%(msecs)03d %(levelname)-8s %(threadName)s %(name)s:%(filename)s:%(lineno)s %(message)s"
+log_date_format = "%Y-%m-%d %H:%M:%S"
```
sveinse, line 12: These are done to replicate HA's setup, yes?

requirements.txt
```diff
@@ -1,10 +1,10 @@
...
+homeassistant==2025.9.4
 pip>=25.0
-ruff==0.12.10
+ruff==0.13.1
 
 # Copy from manifest.json to get this into the dev container
 # without needing to start HA
 azure-servicebus==7.14.2
-pydantic==2.11.7
+pydantic==2.11.9
```
sveinse, line 9: OK. But this is mixing a little bit between "move old API test into pytest framework" and update of external prerequisites if they aren't necessary for doing what the title describes. In an ideal world this should be two PRs. But it's ok.

tests/conftest.py
```diff
@@ -1 +1,51 @@
...
+import os
+
+import pytest
+
+
+@pytest.fixture(scope="session")
+def skip_if_in_github_actions() -> None:
+    """Check if we are running in Github actions and skip any dependant tests if true."""
+    if os.getenv("GITHUB_ACTIONS") == "true":
+        pytest.skip("This test doesn't work in Github Actions.")
```
sveinse, line 12: This is also a good way to do the skip.

tests/zaptec/test_diagnostics.py
sveinse: I think that solution is ok. As you've probably seen already, I posted a proposal to how to do it by the use of additional command line parameters. You can chose which to use. In pytest there isn't a right way and a wrong way, just alternative ways to do it.

pyproject.toml
steinmn: Yes: https://github.com/home-assistant/core/blob/dev/pyproject.toml#L449-L450

tests/zaptec/test_diagnostics.py
steinmn: Considering how limited our testing still is, I think I prefer to have the API-login-dependent tests run by default and then make exceptions where they don't run. Once we have higher test coverage and the ability to mock out stuff requiring login, we can reconsider to only run these full tests when adding additional arguments.

requirements.txt
steinmn: Yeah, I kind of forgot about this one and it got added by accident. Then again, it's only relevant for the dev-container, and it's nice to not let that get too outdated.

sveinse (approved):
I don't see any more objections. Approved.

PR #331, opened 2025-09-28
Release v0.8.3

sveinse:
Note that this also deletes CHANGELOG.md, which we don't really need when making releases in GitHub


sveinse:
@steinmn Please review. Do you think #324 should be included or not? Unless that PR is complete and ready very soon, I'd say we don't wait. Getting official v0.8.3 out is more critical I think.

steinmn (approved):
LGTM

324 would be nice to have, but it also covers a very limited edge case. I'd maybe give them until tomorrow evening to respond and if there is no response by 20:00 or so, just push 0.8.3 without it (assuming you have time to do that tomorrow).

Getting 0.8.3 out before HA 2025.10.0 is definitely more important than 324.

steinmn:
Looks good :+1: 

Maybe in the future, we can create a next_release.md-document in the repo as a draft for the upcoming release notes?

steinmn:
...aaaand I just read your comment on https://github.com/custom-components/zaptec/pull/268#issuecomment-3215067716 :facepalm: You should probably add that too :rofl:

sveinse:
What the hell. I'm a goldfish 🐟

sveinse:
Being spread a little too thin and slipping the details tells me two things: a) here and now fix is much more valuable than "lets remember to note it later", b) A PR puts a record in the git history which is traceable and will make it to the changelog. List of issues is not. So its important to get an explicit PR for something new. "Explicit is better than implicit." [Zen of python](https://peps.python.org/pep-0020/)

steinmn:
Completely forgot about that one myself, so you're in good company in the aquarium :stuck_out_tongue:

And in the spirit of not trusting myself to remember this tomorrow: https://github.com/custom-components/zaptec/pull/334

PR #334, opened 2025-09-29
Add CHANGELOG.md for keeping important release notes

steinmn:
so we don't have to trust our memory


next_release.md
```diff
@@ -0,0 +1,14 @@
...
+It also allows us to include discussions on the wording of release notes in a PR before the final release notes are published.
+
+Once the new version is released, this document should be reset and we start gathering topics for the next version after that.
+
+# v0.8.4
+
+## 🎉 What's in this release
+
+## 🌻 What's Changed
+-
```
sveinse: The _What's in this release_ section is the important section, because this is always manually written. It should be the section that highlight the release, so it makes sense to review this to make sure its good.

The _What's Changed_ section, however, is automatically generated and can be omitted from this file I think.

steinmn: Wasn't sure what was automatic and what was semi-automatic/generated-with-command-or-script, so added it just in case. Removed now.

steinmn: removed next_release in favor of reinstating a modified CHANGELOG.md

PR #338, opened 2025-10-08
Update pydantic to 2.12

### LINKED ISSUE #337: Update pydantic to 2.12, actually perform validation, and prepare for 0.8.4-release
https://github.com/home-assistant/core/pull/153937 updated the pydantic version on the dev-branch, so we will need to follow suit. Will probably be released in 2025.11.0, but from the last run-in with this, we can expect issues from 2025.11.0b0 beta-testers as well, so we should probably be ready before that drops.

Need to test if we trigger any incompatibilities by switching from 2.11 to 2.12 to determine if we can just relax our pydantic-requirement, or if we need to increase the minimum HA-version.

steinmn:
Fixes #337 
- [x] Step 1: Ensure full test coverage of validate.py (when including login credentials)
- [x] Step 2: Update Pydantic to 2.12
- [x] Step 3: Fix breaking changes
- [x] Step 4: Expand test_validate.py to include more real(ish) input data
- [x] Step 5: Expand documentation in comments and/or DEVELOPMENT.md

### DIFF
--- a/custom_components/zaptec/zaptec/validate.py
+++ b/custom_components/zaptec/zaptec/validate.py
@@ -49,11 +49,6 @@ class ChargerState(BaseModel):
     # ValueAsString: str # StateId -1 (Pulse) does not include a ValueAsString-field  # noqa: E501, ERA001
 
 
-ChargerStates = TypeAdapter(list[ChargerState])
-ChargerUpdate = TypeAdapter(dict[str, str])
-Constants = TypeAdapter(dict[str, Any])
-
-
 class Chargers(BaseModel):
     """Pydantic model for a list of Zaptec chargers."""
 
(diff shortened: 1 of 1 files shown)


custom_components/zaptec/zaptec/validate.py
```diff
@@ -123,22 +116,30 @@ class InstallationConnectionDetails(BaseModel):
     Topic: str
 
 
+CHARGER_FIRMWARES = TypeAdapter(list[ChargerFirmware])
+CHARGER_STATES = TypeAdapter(list[ChargerState])
+CHARGER_UPDATES = TypeAdapter(dict[str, str])
+CONSTANTS = TypeAdapter(dict[str, Any])
```
sveinse, line 122: Why this change? Because of the lint recommendation that top-level variables are named in upper case, or something else?

steinmn: Given that it's now an instance instead of a class, PascalCase didn't seem right to me, and it also seemed like a good idea to have a visual distinction between the two in the URLS-map. I don't remember if ruff complained about this, really just a personal preference, and I probably won't be hard to convince if you prefer a different case.

sveinse: It's a good practice to do so, but the reasoning to why weren't immediately clear to me. I'm good as this is now.

PR #339, opened 2025-10-10
Fix bug in state redaction and add full test coverage for redact.py

steinmn:
Bug description:
The key input to the redaction became "950 (MacMain)", which did not match the REDACT_KEYS entry "MacMain", so the value was not redacted.

### DIFF
--- a/custom_components/zaptec/zaptec/redact.py
+++ b/custom_components/zaptec/zaptec/redact.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from collections.abc import Iterable
+from collections.abc import Sequence
 from pprint import pformat
 from typing import Any, ClassVar, TypeVar, cast
 
@@ -159,18 +159,20 @@ class Redactor:
         return obj
 
     def redact_statelist(
-        self, objs: Iterable[dict[str, str]], ctx: str = ""
-    ) -> Iterable[dict[str, str]]:
+        self, objs: Sequence[dict[str, str]], ctx: str = ""
+    ) -> Sequence[dict[str, str]]:
         """Redact the special state list objects."""
         for obj in objs:
             for key in self.OBS_KEYS:
                 if key not in obj:
                     continue
-                keyv = ZCONST.observations.get(obj[key])
-                if keyv is not None:
+
+                # Get the string for the observation key
+                keyv: str = ZCONST.observations.get(obj[key], "")
+                if keyv:
                     obj[key] = f"{obj[key]} ({keyv})"
-                if keyv not in self.REDACT_KEYS:
-                    continue
+
+                # Redact the value if needed
                 for value in self.VALUES:
                     if value not in obj:
                         continue


custom_components/zaptec/zaptec/redact.py
```diff
@@ -165,5 +165,5 @@ def redact_statelist(self, objs, ctx=None):
                 for value in self.VALUES:
                     if value not in obj:
                         continue
-                    obj[value] = self(obj[value], key=obj[key], ctx=ctx)
+                    obj[value] = self(obj[value], key=keyv, ctx=ctx)
```
sveinse, line 179: Good catch on this issue. However, I don't think we have a complete fix for the issue. `keyv` might be `None` if the `obj[key]` is not listed in `ZCONST.observations` [^1], L160. So what should the `key=` argument be when `keyv` is `None`? Perhaps:

```py
obj[value] = self(obj[value], key=keyv or "", ctx=ctx)
````

`key` is kind of useless without a suitable string in `keyv`.

[^1]: I really dislike that the `ZCONST.observations` is s two-way lookup; int->str and str->int. It's a nightmare for type annotations and it should be refactored With this setup `keyv` _might_ also return an `int` if `obj[key]` happens to be a `str`.

Comment (steinmn) [reply custom_components/zaptec/zaptec/redact.py:179]: Isn't that case caught by L163? Or does `None not in self.REDACT_KEYS` evaluate to false? Definitely could be made more explicit though...

**Comment by steinmn** [comment]
> May I suggest that we merge these two into this PR?

Sure, maybe merge the pydantic-update into master first so we can rebase on that and not have the validate-manifest-test failing?

File: custom_components/zaptec/zaptec/redact.py (hunk 1 above)
Comment (sveinse) [reply custom_components/zaptec/zaptec/redact.py:179]: Yes, you're right. This also means that if the `keyv` is not in `self.REDACT_KEYS` the value will never be checked for redactable content, which is possibly an oversight. Perhaps add a test for that case?

File: custom_components/zaptec/zaptec/redact.py [hunk 2]
```diff
@@ -115,15 +118,15 @@ def __call__(self, obj: T, *, key=None, second_pass=False, ctx=None) -> T:
         if isinstance(obj, (tuple, list, set)):
             # Redact each element in the list
             return cast(
-                T,
+                "T",
```
steinmn: Found that Core has disabled TC006 (https://github.com/home-assistant/core/blob/dev/pyproject.toml#L777-L778), so adding the quotes is now optional.

sveinse: Yes, then this quoting can be removed.

custom_components/zaptec/zaptec/redact.py
sveinse: I think maybe the code can be something like this:

```py
        for obj in objs:
            for key in self.OBS_KEYS:
                if key not in obj:
                    continue

                # Get the string for the observation key
                keyv: str = ZCONST.observations.get(obj[key], "")
                if keyv:
                    # Write the observation string name into the state object
                    obj[key] = f"{obj[key]} ({keyv})"

                # Redact the value if needed
                for value in self.VALUES:
                    if value not in obj:
                        continue
                    obj[value] = self(obj[value], key=keyv, ctx=ctx)
```

Note that I haven't tested this. A test case should be written to verify the intended behavior. 

This code yields an type checking error because of `ZCONST.observations` dual type. I'm not sure right now how it's best to deal with that duality without setting up more run-time checks.

steinmn: > A test case should be written to verify the intended behavior.

Done, and code works as intended. Not often I get to do actual test-driven-development and write the test before the fix, but every time I do, it feels really nice.

Sidenote: Apparently you don't get the `__getitem__`-function until you get to Sequence, so I swapped Iterable for that in order to avoid lots of type checking errors in the test file.

steinmn: > This code yields an type checking error because of `ZCONST.observations` dual type. I'm not sure right now how it's best to deal with that duality without setting up more run-time checks.

Create a new issue called "Refactor ZCONST.observations" where you describe the problems with two-way lookup/typing?

custom_components/zaptec/zaptec/redact.py
steinmn: Done

custom_components/zaptec/zaptec/redact.py
sveinse: Done. See #341 and #342

Issue #343, opened 2025-10-14
Sensors unavailable after update and back


marcbaul:
After an update to 0.8.3 and reverse to 0.8.2 the sensors in Zaptec are unavailable. In the log the following error appear:

`Logger: homeassistant.components.sensor
Source: helpers/entity_platform.py:451
integration: Sensor (documentation, issues)
First occurred: 8:03:09 AM (1 occurrence)
Last logged: 8:03:09 AM

Error while setting up zaptec platform for sensor: 'str' object has no attribute 'get'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 451, in _async_setup_platform
    await asyncio.shield(awaitable)
  File "/config/custom_components/zaptec/sensor.py", line 354, in async_setup_entry
    entities = entry.runtime_data.create_entities_from_zaptec(
        INSTALLATION_ENTITIES,
        CHARGER_ENTITIES,
    )
  File "/config/custom_components/zaptec/__init__.py", line 478, in create_entities_from_zaptec
    self.create_entities_from_descriptions(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        charger_descriptions,
        ^^^^^^^^^^^^^^^^^^^^^
        obj,
        ^^^^
        info,
        ^^^^^
    )
    ^
  File "/config/custom_components/zaptec/__init__.py", line 422, in create_entities_from_descriptions
    updater()
    ~~~~~~~^^
  File "/config/custom_components/zaptec/sensor.py", line 121, in _update_from_zaptec
    session_reading = get_ocmf_max_reader_value(session.get("SignedSession", {}))
                                                ^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'get'`

and after

`This error originated from a custom integration.

Logger: custom_components.zaptec.api
Source: custom_components/zaptec/api.py:168
integration: Zaptec EV charger (documentation, issues)
First occurred: 8:03:09 AM (2 occurrences)
Last logged: 8:03:10 AM

[...]

PR #344, opened 2025-10-14
Do not call get on completed_session unless it's a dict

steinmn:
Does not fix the underlying problem of signedsession being null in #343, but at least it doesn't crash the sensor setup.


steinmn:
https://github.com/custom-components/zaptec/actions/runs/18541493246/job/52849850821 reproduces the error message by including `"SignedSession":null` in the input to the complete_session-type-conversion. Still not sure why it is null, but in the description of the completed_session-object in https://docs.zaptec.com/reference/api_chargehistory_get, SignedSession is listed as `string | null`, so I guess this is expected behaviour?

PR #352, opened 2025-11-04
Fix incorrect validation model for charger update

### LINKED ISSUE #351: Failed to validate chargers/xxxxxxx-bc0c-4a92-b0f0-xxxxxx/update
It seems latest **0.8.4** has some issue with my charger.
Trying to exec commands in HA they fail (partly) throwing the error below in the logs.

I rolled back to previous 0.8.3 and it seem to work as expected.

```
2025-11-04 19:57:28.277 ERROR (MainThread) [custom_components.zaptec.zaptec.validate] Failed to validate chargers/xxxxxxx-bc0c-4a92-b0f0-xxxxxx/update (pattern chargers/[0-9a-f\-]+/update): 1 validation error for dict[str,str]
  Input should be a valid dictionary [type=dict_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/dict_type
```

I replaced some chars in the guid tho my guid matches the regex.

Image

steinmn:
api/chargers/{id}/update only returns an empty response (with a 200/401/403 response code). Not sure if this is due to a change in the API or if it's just a sleeper-bug that was introduced in #183 and never triggered because the validation code was not executed previously.

Fixes #351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.8.5
+
+This is a minor release that fixes a bug in the setting of the Min/Max current of a charger.
+
+This update is required to be able to run on HA Core 2025.11+.
+
 ## 0.8.4
 
 This is a minor release that is required to be able to run on HA Core 2025.11+.


sveinse:
Does the code handle getting empty responses to update?

steinmn:
Included a reference to the `HA Core 2025.11+`-compatibility in the changelog entry, since we may need to push this out rather quickly. I think setting the min/max current values on the charger with 0.8.4 will work, it's only the return response that triggers an error (but I think that's bad enough).

steinmn:
> I'm happy to push a 0.8.5 tonight if you are up for prepping the release PR.

https://github.com/custom-components/zaptec/pull/353

PR #355, opened 2025-11-06
Remove incorrect or too restrictive validations

### LINKED ISSUE #354: Unable to change settings due to too strict validation checks
I'm experiencing a problem with the Zaptec integration in Home Assistant. I'm no longer able to change any settings through Home Assistant. adjustments I make (such as charging current or access control) don’t seem to have any effect. The settings remain unchanged.

This issue appears to be limited to Home Assistant; the Zaptec app and portal still work as expected.

Could you please advise on what might be causing this and how to resolve it?

### LINKED ISSUE #357: Cannot connect when not being owner of charger
Hi,

Since a few days I cannot connect to Zaptec anymore:

Image

I checked my credentials many times, tried to download another version, but nothing changed. Many things changed at the same time (component update to 0.8.5, HAOS update to 2025.11, and security update on Zaptec side). To be honest I can not says which one may be involved.

Error log:

Enregistreur: custom_components.zaptec.zaptec.validate
Source: custom_components/zaptec/zaptec/validate.py:167
intégration: Zaptec EV charger ([documentation](https://github.com/custom-components/zaptec), [problèmes](https://github.com/custom-components/zaptec/issues))
S'est produit pour la première fois: 08:36:54 (1 occurrence)
Dernier enregistrement: 08:36:54

Failed to validate installation (pattern installation): 1 validation error for Installations Data.0.AuthenticationType Field required [type=missing, input_value={'Id': '5cc6e035-ad6b-45d...utomaticUpdates': False}, input_type=dict] For further information visit https://errors.pydantic.dev/2.12/v/missing

steinmn:
Validation of ChargerLocalSettings-response:
We don't use the response for anything, and it is an undocumented API, so we shouldn't require any validation.

Validation of Installation(s)-response:
Users without the Owner role don't get the full installation object, specifically not the AuthenticationType-entry that was in the validation. Since the Installations-response includes all installations the user has access to, not just the one(s) they have added to HA, that causes issues. Since we don't use Authentication type for anything (other than displaying it in a sensor), we can just remove the requirement.

Fixes #354. Fixes #357

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## 0.8.6 (draft)
 
-- Fixes issue where calls to the localSettings-API triggered an error on validating the response
+This is a minor release that fixes a validation issue that caused the integration to stop working for users with at least one charger they only have the User role for.
+
+In addition, it also fixes an issue where calls to the localSettings-API triggered an error on validating the response.
 
 ## 0.8.5
 
--- a/custom_components/zaptec/zaptec/validate.py
+++ b/custom_components/zaptec/zaptec/validate.py
@@ -17,7 +17,6 @@ class Installation(BaseModel):
     model_config = ConfigDict(extra="allow")
     Id: str
     Active: bool
-    AuthenticationType: int
     CurrentUserRoles: int
     InstallationType: int
     NetworkType: int


steinmn:
Added validation-relaxing for Installation to address #357 as well. I think this one is severe enough for another release. Do you think we need a more thorough review of the validation models to see if we really need the values we have set as required @sveinse?

sveinse (approved):
LGTM

But as mentioned, we shouldn't validate anything except the data we use in the integration. I.e. the point of validation is to prevent error deeper in the integration, not check _all_ of Zaptec's data.

PR #356, opened 2025-11-06
More ruff and warning cleanup

steinmn:
Bringing the number of ruff errors down below 50.

Part of #258

### DIFF
--- a/custom_components/zaptec/binary_sensor.py
+++ b/custom_components/zaptec/binary_sensor.py
@@ -12,11 +12,10 @@ from homeassistant.components.binary_sensor import (
     BinarySensorEntityDescription,
 )
 from homeassistant.core import HomeAssistant, callback
-from homeassistant.helpers.entity import EntityDescription
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 
 from .entity import ZaptecBaseEntity
-from .manager import ZaptecConfigEntry
+from .manager import ZaptecConfigEntry, ZaptecEntityDescription
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -44,13 +43,13 @@ class ZaptecBinarySensorWithAttrs(ZaptecBinarySensor):
 
 
 @dataclass(frozen=True, kw_only=True)
-class ZapBinarySensorEntityDescription(BinarySensorEntityDescription):
+class ZapBinarySensorEntityDescription(ZaptecEntityDescription, BinarySensorEntityDescription):
     """Class describing Zaptec binary sensor entities."""
 
     cls: type[BinarySensorEntity]
 
 
-INSTALLATION_ENTITIES: list[EntityDescription] = [
+INSTALLATION_ENTITIES: list[ZaptecEntityDescription] = [
     ZapBinarySensorEntityDescription(
         key="active",
         name="Installation",  # Special case, no translation
@@ -72,7 +71,7 @@ INSTALLATION_ENTITIES: list[EntityDescription] = [
     ),
 ]
 
-CHARGER_ENTITIES: list[EntityDescription] = [
+CHARGER_ENTITIES: list[ZaptecEntityDescription] = [
     ZapBinarySensorEntityDescription(
         key="active",
         name="Charger",  # Special case, no translation
--- a/custom_components/zaptec/diagnostics.py
+++ b/custom_components/zaptec/diagnostics.py
@@ -30,6 +30,7 @@ async def async_get_config_entry_diagnostics(
         return await _get_diagnostics(hass, config_entry)
     except Exception:
         _LOGGER.exception("Error getting diagnostics")
+        return {}
 
 
 async def async_get_device_diagnostics(
@@ -40,6 +41,7 @@ async def async_get_device_diagnostics(
         return await _get_diagnostics(hass, config_entry)
     except Exception:
         _LOGGER.exception("Error getting diagnostics for device %s", device.id)
+        return {}
 
 
 async def _get_diagnostics(
@@ -80,23 +82,24 @@ async def _get_diagnostics(
     try:
         api = out.setdefault("api", {})
 
-        async def request(url: str) -> Any:
+        async def request(url: str) -> dict | list:
             """Make an API request and return the result."""
             try:
                 result = await zaptec.request(url)
                 if not isinstance(result, (dict, list)):
                     return {
-                        "type error": f"Expected dict, got type {type(result).__name__}, value {result}",
+                        "type error": f"Expected dict or list, got type {type(result).__name__}, value {result}",  # noqa: E501
                     }
-                return result
             except Exception as err:
                 return {
                     "exception": type(err).__name__,
                     "err": str(err),
                     "tb": list(traceback.format_exc().splitlines()),
                 }
+            else:
+                return result
 
-        def add(url, obj, ctx=None) -> None:
+        def add(url: str, obj: dict | list, ctx: str = "") -> None:
             api[redact(url, ctx=ctx)] = redact(obj, ctx=ctx)
 
         data = await request(url := "installation")
@@ -109,8 +112,9 @@ async def _get_diagnostics(
 
             for circuit in data.get("Circuits", []):
                 add(f"circuits/{circuit['Id']}", circuit, ctx="circuit")
-                for charger in circuit.get("Chargers", []):
-                    charger_in_circuits_ids.append(charger["Id"])
+                charger_in_circuits_ids.extend(
+                    charger["Id"] for charger in circuit.get("Chargers", [])
+                )
 
             add(url, data, ctx="hierarchy")
 
@@ -126,7 +130,7 @@ async def _get_diagnostics(
             add(url, data, ctx="charger")
 
             data = await request(url := f"chargers/{charger_id}/state")
-            redact.redact_statelist(data, ctx="state")
+            data = redact.redact_statelist(data, ctx="state")
             add(url, data, ctx="state")
 
     except Exception as err:
--- a/custom_components/zaptec/diagnostics.py
+++ b/custom_components/zaptec/diagnostics.py
@@ -88,7 +88,7 @@ async def _get_diagnostics(
                 result = await zaptec.request(url)
                 if not isinstance(result, (dict, list)):
                     return {
-                        "type error": f"Expected dict or list, got type {type(result).__name__}, value {result}",  # noqa: E501
+                        "type error": f"Expected dict or list, got type {type(result).__name__}, value {result}",
                     }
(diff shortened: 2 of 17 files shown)


custom_components/zaptec/zaptec/api.py
```diff
@@ -96,12 +97,12 @@ def __iter__(self) -> Iterator[str]:
     @property
     def id(self) -> str:
         """Return the id of the object."""
-        return self._attrs["id"]
+        return str(self._attrs["id"])
```
sveinse, line 100: Can `self._attrs["id"]` ever be anything else than `str`? While I understand this helps with the type, it will return the string "None" if the object is `None`. Should it rather raise an exception if the object is non-str? -- I'm not sure what's best here, so I'm open to a discussion.

custom_components/zaptec/zaptec/api.py
```diff
@@ -96,12 +97,12 @@ def __iter__(self) -> Iterator[str]:
...
 
     @property
     def name(self) -> str:
         """Return the name of the object."""
-        return self._attrs["name"]
+        return str(self._attrs["name"])
```
sveinse, line 105: Ditto.

custom_components/zaptec/zaptec/api.py
sveinse: We should except this ruff error (D102) for this line. The protocol using ellipsis is correct.

custom_components/zaptec/binary_sensor.py
```diff
@@ -44,13 +43,13 @@ def _post_init(self) -> None:
 
 
 @dataclass(frozen=True, kw_only=True)
-class ZapBinarySensorEntityDescription(BinarySensorEntityDescription):
+class ZapBinarySensorEntityDescription(ZaptecEntityDescription, BinarySensorEntityDescription):
```
sveinse, line 46: If I'm not mistaken, doesn't this result in a diamond inheritance? Does this have any practical effects in this usage? I've been taught to avoid them like the plague.

custom_components/zaptec/diagnostics.py
```diff
@@ -30,6 +30,7 @@ async def async_get_config_entry_diagnostics(
         return await _get_diagnostics(hass, config_entry)
     except Exception:
         _LOGGER.exception("Error getting diagnostics")
+        return {}
```
sveinse, line 33: What's correct behavior here? Is it to return nothing like everything is ok, or would a more correct approach be to escalate the error to HA? What does other integrations do?

custom_components/zaptec/diagnostics.py
```diff
@@ -80,23 +82,24 @@ def add_failure(err: Exception) -> None:
...
                     }
-                return result
             except Exception as err:
                 return {
                     "exception": type(err).__name__,
                     "err": str(err),
                     "tb": list(traceback.format_exc().splitlines()),
                 }
+            else:
+                return result
```
sveinse, line 100: Why this change? I find the way it was more readable than this change. Is there a formatting recommendation behind this one?

custom_components/zaptec/services.py
```diff
@@ -127,7 +127,7 @@ def get_as_set(service_call: ServiceCall, key: str) -> set[str]:
 
     def iter_objects(
         service_call: ServiceCall, mustbe: type[T]
-    ) -> Generator[tuple[ZaptecUpdateCoordinator, T], None, None]:
+    ) -> Generator[tuple[ZaptecUpdateCoordinator, T]]:
```
sveinse, line 130: Because it's a fact that this code will always be run on py3.13+, yes?

custom_components/zaptec/diagnostics.py
```diff
@@ -126,7 +130,7 @@ def add(url, obj, ctx=None) -> None:
             add(url, data, ctx="charger")
 
             data = await request(url := f"chargers/{charger_id}/state")
-            redact.redact_statelist(data, ctx="state")
+            data = redact.redact_statelist(data, ctx="state")
```
sveinse, line 133: This is a good catch. Because `redact_statelist` is currently implicitly modifying `data`. But later implementations might not, so doing this is good practice.

custom_components/zaptec/zaptec/api.py
sveinse, line 302: I suggest excluding the TODO or FIXME linting comments. Or do you think they should be there as a constant reminder that something isn't complete?

custom_components/zaptec/zaptec/api.py
```diff
@@ -849,9 +843,9 @@ async def __aexit__(self, exc_type, exc_value, traceback):
     # =======================================================================
     #   MAPPING METHODS
 
-    def __getitem__(self, id: str) -> ZaptecBase:
+    def __getitem__(self, obj_id: str) -> ZaptecBase:
         """Get an object data by id."""
-        return self._map[id]
+        return self._map[obj_id]
```
sveinse, line 848: /pedantic: I think the standard argument for `__getitem__` is `item`. Since that we use for all the other non-dunder methods I think this is fine.

custom_components/zaptec/zaptec/api.py
```diff
@@ -924,7 +919,7 @@ def _request_log(url, method, iteration, **kwargs):
                 # Remove the Authorization header from the log
                 if "Authorization" in headers:
                     headers["Authorization"] = "<Removed for security>"
-                yield f"     headers {dict((k, v) for k, v in headers.items())}"
+                yield f"     headers {dict(headers.items())}"
```
sveinse: Which is the same as `headers.copy()` isn't it? Which in turn is the same as just doing `{headers}`?

I don't remember why this construct is like this. I suspect there used to be a `if` to that dict comprehension that has been removed.

sveinse (commented):
Great work!. I have a few comments it would be good to discuss

sveinse:
@steinmn Do we want this in place before the pending 0.8.6 release?

steinmn:
> @steinmn Do we want this in place before the pending 0.8.6 release?

I don't think so. It's a lot of small changes which (hopefully) only impacts code quality stuff, so probably better to just get a stable version out and then have some more time to discuss all the points above.

custom_components/zaptec/services.py
steinmn: Pretty sure HA dropped python3.12 before the 2025.7 version we list as minimum, but I'll double check once I have a bit more time to devote.

custom_components/zaptec/diagnostics.py
steinmn: I'm inclined to agree, but https://docs.astral.sh/ruff/rules/try-consider-else/ doesn't, and it does seem to be a convention that is in use in core :shrug:

custom_components/zaptec/zaptec/api.py
steinmn: Can the object be `None`? `_attrs`is a TDict, where the values can be `str | int | float | bool`, so I don't think `None` should be an option? And if the key doesn't exist, then the code will raise a `KeyError`, right?

custom_components/zaptec/binary_sensor.py
steinmn: I'll be honest, I'd never heard of diamond inheritance before. Having said that, is this really any different from
```class ZaptecBinarySensor(ZaptecBaseEntity, BinarySensorEntity):```
a couple of lines above this? Both inherit from the Entity class (if you dig down a bit). Also, these particular classes only add (non-overlapping) variables, so there's no ambiguity.

This was done to avoid the warning `Argument of type "Iterable[EntityDescription]" cannot be assigned to parameter "descriptions" of type "Iterable[ZaptecEntityDescription]" in function "create_entities_from_descriptions"`. Unsure about the practical effects.

File: custom_components/zaptec/diagnostics.py (hunk 4 above)
Comment (steinmn) [reply custom_components/zaptec/diagnostics.py:33]: Looking a bit more thoroughly at it, I don't see how we can ever reach this line of code anyway. Pretty much all of `_get_diagnostics` is try-excepted, so it would have to fail on `add_failure` I think? Part of me wants to delete the entire try-except here, and the more risk-averse part of me wants to leave it as-is.

File: custom_components/zaptec/services.py (hunk 6 above)
Comment (steinmn) [reply custom_components/zaptec/services.py:130]: Python3.12 was dropped in 2025.2.0 [commit](https://github.com/home-assistant/core/commit/60bdc13c9448e60ea257e7281ed4e78ac0f96d4b), so we're good.

File: custom_components/zaptec/zaptec/api.py (hunk 8 above)
Comment (steinmn) [reply custom_components/zaptec/zaptec/api.py:848]: [Here](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) it looks like it's `key`. I think I prefer obj_id to both alternatives.

File: custom_components/zaptec/zaptec/api.py
Comment (steinmn) [reply custom_components/zaptec/zaptec/api.py:302]: Well, all the TODO or FIXME linting comments are about this single line, and I'm not sure what it actually means, what needs to be fixed here? Should the FIXME be replaced with a NOTE instead?

Comment (steinmn) [reply custom_components/zaptec/zaptec/api.py]: :heavy_check_mark:

File: custom_components/zaptec/zaptec/api.py (hunk 9 above)
Comment (steinmn) [reply custom_components/zaptec/zaptec/api.py]: The simpler the better :+1:

File: custom_components/zaptec/zaptec/api.py
Comment (steinmn) [reply custom_components/zaptec/zaptec/api.py:302]: If it is indeed something that needs to be fixed, I think disabling the FIX002-error is fine, but I kind of agree with the other 3 (requiring an author, an issue link to the issue where it will be fixed, and for the keyword to be TODO).

So for instance, this will not give any ruff errors:
```
# TODO(@sveinse): #258 SupportGroup is sub dict. This is not within the declared type  # noqa: FIX002
```

custom_components/zaptec/zaptec/api.py
sveinse: I'm happy with this.

sveinse:
This is a long standing PR and I'm inclined to just get it in there and see what happens. I do not have any detailed objections to it. Any objections to merging this into master at this point? We can probably observe the fallout as we go, right?

custom_components/zaptec/zaptec/api.py
sveinse: I'm not sure what's best atm. So I think we can leave it for later.

sveinse (approved):
LGTM - this takes us closer to a better linted codebase, while not 100% perfect, its good enough to move us forward.

steinmn:
> This is a long standing PR and I'm inclined to just get it in there and see what happens. I do not have any detailed objections to it. Any objections to merging this into master at this point? We can probably observe the fallout as we go, right?

Agreed 👍

PR #367, opened 2026-02-26
Added a note in the documentation regarding stand-alone mode

no-gram:
Added a note in the documentation regarding stand-alone mode in the section about prevent charging auto start

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,12 @@ of the following will prevent auto start:
 > Using _"Available current"_ will affect __all__ chargers if there are more
 > than one charger.
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


no-gram:
When stand-alone mode is activated the integration works as expected. The charging started immediately and the state was correctly updated in HA. I could turn off the charging through the integration and I was considering this as a workaround now that I could not prevent the auto start. Stand-alone is supported by the integration and as far as I see it, it works as expected. It should not be a requirement that stand-alone mode is turned off.

steinmn (approved):
Good catch on the dead link 👍

sveinse (approved):

sveinse:
@no-gram Thank you so much for the contribution. PR merged.

PR #387, opened 2026-07-06
Fix reportIncompatibleVariableOverride type errors in platform entities

rhammen:
## Problem

The entity platform modules (`switch.py`, `sensor.py`, `number.py`, `button.py`, `binary_sensor.py`, `update.py`) produced **31 near-identical `reportIncompatibleVariableOverride` errors** in Pylance/pyright (default `standard` mode). They came in three repeating shapes:

1. **`zaptec_obj` narrowing (8):** `ZaptecBaseEntity` declares `zaptec_obj: ZaptecBase`; subclasses redeclared it as `Charger`/`Installation`. Because a plain class attribute is mutable, its type is invariant, so narrowing it in a subclass is flagged as an incompatible override.
2. **`entity_description` diamond + narrowing (11):** `ZaptecBaseEntity` redeclared `entity_description: EntityDescription`, which conflicts with the HA platform mixin's own narrower `entity_description` (e.g. `SwitchEntity.entity_description: SwitchEntityDescription`) in the `class ZaptecX(ZaptecBaseEntity, HAEntity)` diamond, and the leaf redeclarations were flagged as narrowing overrides too.
3. **`cls` narrowing (6):** each `ZapXEntityDescription` redeclared `cls: type[SwitchEntity]` / `type[SensorEntity]` / etc. Those types are **not** subtypes of the base's `cls: type[ZaptecBaseEntity]`, so every one was an incompatible (and actually incorrect) override.

These were pure typing errors — the integration ran fine — but they cluttered the editor with red squiggles across every platform file.

## What changed and why

**1. Make `ZaptecBaseEntity` generic over the Zaptec object type.**
`class ZaptecBaseEntity[ZaptecObjectT: ZaptecBase = ZaptecBase]` with `zaptec_obj: ZaptecObjectT`. Subclasses now narrow by parametrizing the base (`ZaptecBaseEntity[Charger]`, `ZaptecNumber[Installation]`, …) instead of redeclaring `zaptec_obj`. This is the idiomatic fix — the same pattern HA core uses for coordinator entities — and removes all 8 `zaptec_obj` errors with **no suppressions**. Generics are erased at runtime, so there is no behavioural change.

**2. Drop the redundant `entity_description: EntityDescription` from `ZaptecBaseEntity`.**
HA's `Entity` already declares `entity_description: EntityDescription`, so the base class inherits it and typing still resolves. Removing our redeclaration clears the base-vs-mixin diamond conflicts and the leaf narrowing errors.

**3. Remove the mistyped `cls: type[XEntity]` redeclarations** in the `ZapXEntityDescription` dataclasses. They now inherit the correct `cls: type[ZaptecBaseEntity[Any]]` from `ZaptecEntityDescription` (widened from `type[ZaptecBaseEntity]` so any generic parametrization is accepted). The `cls` dataclass field is still present (verified at runtime).

### Result

Platform-file errors: **31 → 6**, zero new errors, zero `# type: ignore`.

The remaining 6 are `available` diamond conflicts (`CoordinatorEntity.available` property vs `Entity.available` cached_property) that are inherent to *every* HA coordinator-based entity and are deliberately left untouched rather than papered over.

## Testing

- `ruff check custom_components tests` — clean on the touched files.
- `ruff format custom_components tests --diff` — clean.
- `pytest tests` (with `SKIP_ZAPTEC_API_TEST=true`) — **10 passed, 2 skipped**; the only errors are the pre-existing `test_zconst.py`/`test_redact.py` cases that require live access to `api.zaptec.com` and are unrelated to this change.
- Runtime smoke check: all platform modules import, MRO is intact, and the `cls` dataclass field is still present after the refactor.

## Notes

No functional/behavioural changes — this is a types-only cleanup. Net diff: 8 files, +17 / -36.

### DIFF
--- a/custom_components/zaptec/binary_sensor.py
+++ b/custom_components/zaptec/binary_sensor.py
@@ -46,8 +46,6 @@ class ZaptecBinarySensorWithAttrs(ZaptecBinarySensor):
 class ZapBinarySensorEntityDescription(ZaptecEntityDescription, BinarySensorEntityDescription):
     """Class describing Zaptec binary sensor entities."""
 
-    cls: type[BinarySensorEntity]
-
 
 INSTALLATION_ENTITIES: list[ZaptecEntityDescription] = [
     ZapBinarySensorEntityDescription(
--- a/custom_components/zaptec/button.py
+++ b/custom_components/zaptec/button.py
@@ -18,10 +18,8 @@ from .zaptec import Charger
 _LOGGER = logging.getLogger(__name__)
 
 
-class ZaptecButton(ZaptecBaseEntity, ButtonEntity):
+class ZaptecButton(ZaptecBaseEntity[Charger], ButtonEntity):
     """Base class for Zaptec buttons."""
-
-    zaptec_obj: Charger
 
     @property
     def available(self) -> bool:
@@ -47,8 +45,6 @@ class ZaptecButton(ZaptecBaseEntity, ButtonEntity):
 @dataclass(frozen=True, kw_only=True)
 class ZapButtonEntityDescription(ZaptecEntityDescription, ButtonEntityDescription):
     """Class describing Zaptec button entities."""
-
-    cls: type[ButtonEntity]
 
 
 INSTALLATION_ENTITIES: list[ZaptecEntityDescription] = []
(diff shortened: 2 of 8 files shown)


sveinse (approved):
I like this PR. Using type parameter syntax makes this so much simpler. Approved.

sveinse:
Thank you for your contributions @rhammen (and Claude :P )

PR #390, opened 2026-07-08
Fix ruff lint errors in api.py

rhammen:
## Summary

Resolves all **36 `ruff check` errors** in the vendored Zaptec API client (`custom_components/zaptec/zaptec/api.py`), taking that file from 36 → 0. This chips away at the non-clean `ruff check` baseline tracked in #258 (CI currently runs `ruff check` with `--exit-zero`).

Almost entirely annotations, comment/logging tidy-ups, and two safe latent-type fixes. See the **Behavior note** below for the one edge case that is not purely cosmetic.

### Lint fixes
- Add missing return-type and argument annotations (`ANN001/003/201/204/205`)
- Replace the mutable default arg for `excludes` with the `None`-default pattern (`B006`)
- Move the "stream already running" precondition `raise` out of the `try` (`TRY301`)
- Bind loop-captured log messages as an immutable default arg (`B023`)
- Reword a stale `FIXME` comment to `NOTE` (`TD001/002/003`, `FIX001`)
- Remove commented-out debug lines (`ERA001`)
- Drop redundant exception args from `logging.exception(...)` calls (`TRY401`)
- Remove an unnecessary `AsyncGenerator` `None` type arg (`UP043`)

### Two safe latent-type fixes (surfaced while annotating)
- Initialize `operation_mode`/`final_stop_active` so they aren't possibly-unbound in `is_command_valid`'s debug log
- Widen `Zaptec.__contains__` key param to `object` for a compatible `Mapping` override

### ⚠️ Behavior note (TRY301)
The `TRY301` fix moves the "stream already running" precondition `raise RuntimeError(...)` from inside `stream_main`'s `try` to before it. Previously that error was caught by the method's own `except Exception` (logged as "Stream failed", and the `finally` reset `_stream_running`); now it propagates to the caller instead. This guard is only hit if `stream_main` is invoked while a stream is already running — which shouldn't happen in normal use (`stream()` calls `cancel_stream()` first, and the manager starts one background task per installation) — and no test exercises it. Arguably more correct (a programming-error guard shouldn't be swallowed by the stream's own handler, and it no longer clobbers a live stream's `_stream_running` flag), but calling it out since it's the one non-cosmetic change. Happy to use a `# noqa: TRY301` instead if maintainers prefer to keep the old swallow-and-log behavior.

### Verification
- `ruff check custom_components tests` — passes (api.py now 0 errors)
- `ruff format --diff` — clean
- Test suite passes (offline subset; the two live-API fixture files are excluded per the known DNS/env gap)

## Follow-up (separate PRs)
This is deliberately scoped to lint-only. I'd like to follow up with staged PRs to raise `api.py` to high-quality code:
1. **Offline unit tests** for the core logic (retry/backoff, request status handling, `state_to_attrs`, `is_command_valid`, `stream_update`, hierarchy parsing) — currently the only api test requires live login.
2. **Type foundation** — annotate instance attrs, parameterize bare generics, `Mapping` covariance, typed accessors to remove `ZaptecBase`→`Charger`/`Installation` downcasts.
3. **Typed API response models** (TypedDict/pydantic) to eliminate the pervasive `Any` on API payloads.
4. **Correctness cleanups** — the `command()` variable reuse + `int(final_stop_active)` guard, explicit kwargs for `set_limit_current`, instance-scoped debug flags.

Happy to split/reorder these however maintainers prefer.

### DIFF
--- a/custom_components/zaptec/zaptec/api.py
+++ b/custom_components/zaptec/zaptec/api.py
@@ -3,14 +3,16 @@
 from __future__ import annotations
 
 import asyncio
-from collections.abc import AsyncGenerator, Callable, Iterable, Iterator, Mapping
+from collections.abc import AsyncGenerator, Awaitable, Callable, Iterable, Iterator, Mapping
 from contextlib import aclosing
 from http import HTTPStatus
 import itertools
 import json
 import logging
 import random
+import ssl
 import time
+from types import TracebackType
 from typing import Any, ClassVar, Protocol, Self
 
 import aiohttp
@@ -59,6 +61,7 @@ DEBUG_API_EXCEPTIONS = False
 # Type definitions
 TValue = str | int | float | bool
 TDict = dict[str, TValue]
+StreamCallback = Callable[[dict], Awaitable[None]]
 
 
 class TLogExc(Protocol):
@@ -186,13 +189,15 @@ class ZaptecBase(Mapping[str, TValue]):
         data: Iterable[dict[str, str]],
         key: str,
         keydict: dict[str, str],
-        excludes: set[str] = set(),
+        excludes: set[str] | None = None,
     ) -> dict[str, str]:
         """Convert a list of state data into a dict of attributes.
 
         `key` is the key that specifies the attribute name. `keydict` is a
         dict that maps the key value to an attribute name.
         """
+        if excludes is None:
+            excludes = set()
         out: dict[str, str] = {}
         for item in data:
             skey = item.get(key)
@@ -299,7 +304,7 @@ class Installation(ZaptecBase):
 
         # Remove data fields with excessive data, making it bigger than the
         # HA database appreciates for the size of attributes.
-        # FIXME: SupportGroup is sub dict. This is not within the declared type
+        # NOTE: SupportGroup is a sub dict. This is not within the declared type
         supportgroup = data.get("SupportGroup")
         if supportgroup is not None and "LogoBase64" in supportgroup:
             logo = supportgroup["LogoBase64"]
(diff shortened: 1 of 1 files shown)


custom_components/zaptec/zaptec/api.py
```diff
@@ -988,12 +1001,17 @@ async def _request_worker(
...
+                    # iteration's values rather than the loop variables.
+                    def log_exc(
+                        exc: Exception,
+                        _log_msgs: tuple[str, ...] = tuple(log_req + log_resp),
+                    ) -> Exception:
                         """Log the exception and return it."""
                         if DEBUG_API_EXCEPTIONS:
                             if not DEBUG_API_CALLS:
-                                for msg in log_req + log_resp:
+                                for msg in _log_msgs:
```
sveinse, line 1014: What is gained from passing this through a default argument? Is there a side-effect that weren't identified earlier? But I like it, because it removes the non-local scope from the callback and only relies on local vars.

sveinse (approved):
I think this looks great.

custom_components/zaptec/zaptec/api.py
rhammen: Good question — nothing new was surfaced, and there's no side-effect in current use. What it fixes is a latent one flagged by ruff's B023.

`log_exc` is defined inside the retry loop, and `log_req`/`log_resp` are rebound on each iteration. Before, the closure read them non-locally, so its output depended on *when* it was called relative to the loop advancing. Since `log_exc` is yielded to the caller, a caller that held onto it and then drove the generator into the next retry (rebinding those vars) would end up logging the wrong iteration's request/response. In practice every caller invokes `log_exc` immediately on the yielded response before retrying, so the values always coincide today — hence no observable change.

Binding them as the `_log_msgs` default arg captures that iteration's values at definition time, so correctness no longer depends on call ordering — which is the non-local-scope removal you mentioned. So: a real fix for a latent capture bug, plus the readability win, with no behavior change for existing callers.

sveinse: I'm really impressed what Claude is capable of doing. I use it for work and find it very useful. What model and effort have you used for these PRs?

rhammen: Yes, I'm also quite impressed 😄 . Used some 'free' AI models earlier, and although those are already really helpfull, they fail at the complex edge cases that Claude catches. 

Most PRs were done with Opus 4.8, some with Sonnet. Still figuring out a bit which model to use when. With Opus and these PRs I ended up using most of my session credits after 2h-3h, so it is still manageable.
Didn't touch the effort, so looks like that was on High all the time.

PR #394, opened 2026-07-11
test: add unit tests for ZaptecUpdateCoordinator and ZaptecBaseEntity

rhammen:
## Summary
- Adds unit tests for `custom_components/zaptec/coordinator.py` (`ZaptecUpdateCoordinator`) and `custom_components/zaptec/entity.py` (`ZaptecBaseEntity`), raising coverage on these files from ~27-29% to 100% and 98% respectively (the one remaining line in `entity.py` is a genuinely unreachable defensive fallback).
- Since neither class can be meaningfully tested with bare mocks (both touch real event-loop scheduling and `ConfigEntry` internals), and the standard `pytest-homeassistant-custom-component` harness is not viable on native Windows dev setups, tests use small hand-rolled test doubles instead: a `MagicMock`-based fake `hass` with a real running event loop attached as `.loop`, and a minimal `FakeConfigEntry` class (both in `tests/conftest.py`). This keeps the tests fast, dependency-free, and runnable everywhere `pytest` runs.
- Along the way, documents (via a test, not a production-code change) a pre-existing gap in `entity.py`: `_handle_coordinator_update` sets `self._attr_available = False` on a `KeyUnavailableError`, but `ZaptecBaseEntity` never overrides HA's `CoordinatorEntity.available` property, so that flag currently has no effect on the entity's actual reported availability. Left as a documented finding for a separate follow-up, out of scope here.

## Test plan
- [x] `pytest tests/test_coordinator.py tests/test_entity.py -v` — 35 passed
- [x] `pytest tests -q` — 45 passed, 2 skipped, 22 errors (pre-existing, unrelated DNS-fixture gap in `test_zconst.py`/`test_redact.py` — see repo's dev notes)
- [x] `ruff format --diff` / `ruff check` clean on all touched files
- [x] No `custom_components/**` production files modified — test-only change


docs/superpowers/plans/2026-07-11-coordinator-entity-tests.md
sveinse, line 1: Please remove this file from the PR.

tests/conftest.py
```diff
@@ -2,12 +2,63 @@
...
 
 
+class FakeConfigEntry:
+    """Minimal stand-in for HA's ConfigEntry.
+
+    Exposes only what coordinator.py and entity.py actually touch
+    (`pref_disable_polling`, `async_on_unload`,
+    `async_create_background_task`). A real ConfigEntry pulls in HA's full
+    test-harness machinery, which cannot run on native Windows in this dev
+    environment - see CLAUDE.md's environment notes.
```
sveinse: I'm not sure about this comment. First about mentioning Windows and second about CLAUDE.md which isn't part of this project.

tests/test_entity.py
```diff
@@ -0,0 +1,310 @@
...
+
+    with caplog.at_level(logging.INFO):
+        entity._handle_coordinator_update()  # noqa: SLF001
+
+    # NOTE: this sets _attr_available, but ZaptecBaseEntity does not override
+    # the `available` property inherited from HA's CoordinatorEntity (which
+    # returns coordinator.last_update_success instead), so this flag currently
+    # has no effect on the entity's actual reported availability. This test
+    # documents today's real behavior, not the intended one - see the "Known
+    # finding" note at the top of this plan.
```
sveinse: Is this something that should be reported as a bug?

The text refers to "known finding" note at the top of this plan. What plan?

sveinse (changes requested):
See the comments for adjustments.

Making unittests for testing HA functionality like coordinator and entities is something they must have a lot of in HA. Have you checked what testing facilities exists in HA that can assist us? How does other components do this type of testing? (And I would look at gold or platinum components in HA core for inspiration.)

I'd also like @steinmn comments on the PR if that's possible. You have a different set of eyes than me.

rhammen:
Thanks — fair point, and having dug into it I agree the current approach doesn't really match how gold/platinum HA integrations test.

Those integrations lean on `pytest-homeassistant-custom-component`: a real `hass`, `MockConfigEntry`, and tests that set the integration up through the normal `async_setup` path with the cloud API mocked, then assert on public state (`hass.states.get(...)`, the entity/device registries) — often via snapshot tests. This PR instead instantiates the coordinator/entity classes directly and asserts against private methods (`# noqa: SLF001` throughout), which couples the tests to internals and tests implementation rather than behavior.

The original reason for the hand-rolled mocks was that the real harness didn't run on my native-Windows dev setup (`homeassistant` imports `fcntl` unconditionally). But that's a local-dev constraint, not a project one — CI runs on Linux where the harness works fine, and the Windows issue is solvable with a small, OS-guarded compat shim.

As a concrete example of why the harness is worth it: the `_attr_available` gap I documented here (now filed as #410) would become a real *failing* assertion under a full-setup test — the entity visibly staying `available` when it shouldn't — instead of a passing test that only checks the internal flag.

So rather than merge this as-is, I'm going to **hold this PR as a draft** and open a replacement built on the `pytest-homeassistant-custom-component` approach. I'll apply the same strategy to #395 (platform-entity tests), which I'm also marking draft for now. Design work is in progress. Would very much welcome @steinmn's eyes on the direction.

steinmn:
> The original reason for the hand-rolled mocks was that the real harness didn't run on my native-Windows dev setup (homeassistant imports fcntl unconditionally). But that's a local-dev constraint, not a project one — CI runs on Linux where the harness works fine, and the Windows issue is solvable with a small, OS-guarded compat shim.

I think I mentioned it on a different PR, not sure which: Have you tried working with the dev-container defined by [.devcontainer.json](https://github.com/custom-components/zaptec/blob/master/.devcontainer.json)? This is exactly the kind of dev-setup-issues this was created to handle.

As long as you have docker running on WSL2 and your repo download location is on the wsl-filesystem (see https://endjin.com/blog/supercharge-dev-containers-on-windows for details on why), you can get just as good performance running in a dev-container as running locally. At my day-job, we recently even managed to reduce the build-time of a c-project by 80% by moving from a native Windows pipeline to a dev-container running on WSL2.

rhammen:
> > The original reason for the hand-rolled mocks was that the real harness didn't run on my native-Windows dev setup (homeassistant imports fcntl unconditionally). But that's a local-dev constraint, not a project one — CI runs on Linux where the harness works fine, and the Windows issue is solvable with a small, OS-guarded compat shim.
> 
> I think I mentioned it on a different PR, not sure which: Have you tried working with the dev-container defined by [.devcontainer.json](https://github.com/custom-components/zaptec/blob/master/.devcontainer.json)? This is exactly the kind of dev-setup-issues this was created to handle.
> 
> As long as you have docker running on WSL2 and your repo download location is on the wsl-filesystem (see https://endjin.com/blog/supercharge-dev-containers-on-windows for details on why), you can get just as good performance running in a dev-container as running locally. At my day-job, we recently even managed to reduce the build-time of a c-project by 80% by moving from a native Windows pipeline to a dev-container running on WSL2.

Yes, I have setup the DEVcontainer this weekend and it is working.
Not sure yet I really like it against my old way of working 😃 , but you are right: it solves the native-Windows constraints, and I shouldn't pollute this repo with workarounds for my personal workflow.
I use the DEVcontainer now at least to running the testsuite, to confirm all tests pass OK.
Thanks for the heads up!

rhammen:
Closing this PR. 
It is superseded by #414, which implements the pytest-homeassistant-custom-component harness which is preferred to gold/platinum code.

PR #395, opened 2026-07-11
test: add entity test coverage for sensor/switch/number/button/binary_sensor/update platforms

rhammen:
## Summary
- Adds unit tests for the six HA entity-platform files that previously had 0% coverage: `sensor.py`, `switch.py`, `number.py`, `button.py`, `binary_sensor.py`, `update.py`.
- Exercises real entity-class logic (value transforms, `_post_init` overrides, `available` overrides, async command methods with success/`HomeAssistantError`-failure paths) using `MagicMock(spec=Charger)`/`MagicMock(spec=Installation)` test doubles so a typo'd method name fails loudly instead of silently returning a fresh mock.
- **Stacked on #394** (adds the `hass`/`config_entry` fixtures and `coordinator.py`/`entity.py` coverage this branch builds on).
- Also includes one commit from a separate, already-reviewed workstream: `services.py` test coverage plus a real `services.yaml` bugfix (documented service key was `update_firmware`, actual registered service is `upgrade_firmware`).

Coverage after this branch: sensor.py 97%, switch.py 97%, number.py 97%, button.py 93%, binary_sensor.py 93%, update.py 94% — remaining gaps are exactly each file's `async_setup_entry()` body (declarative/thin-delegation code, out of scope by design).

## Test plan
- [x] `SKIP_ZAPTEC_API_TEST=true pytest tests -q` → 132 passed, 2 skipped, 22 errors (pre-existing DNS-fixture gap in `test_zconst.py`/`test_redact.py`, documented in `CLAUDE.md`, unrelated to this change)
- [x] `ruff format --diff` / `ruff check` clean on all new files
- [x] Each task individually reviewed (spec compliance + code quality, both approved) plus a final whole-branch review (no Critical/Important findings)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #393, opened 2026-07-09
Retry transient server errors on setup (fixes #392)

rhammen:
Fixes #392. Closes #360.

> Part A (retrying transient `5xx` for the token request) is exactly what #360 ("Retry token on 503 responses") asked for, so this PR resolves both.

Built on top of #391 (`Add offline unit tests for the Zaptec API client`) — the tests here reuse its `FakeSession` / `_make_zaptec` harness.

## The bug

On a fresh install (e.g. a dev-container), adding a Zaptec account intermittently fails setup with a `503` from the token endpoint, and only a full Home Assistant restart recovers:

```
RequestError: POST request to https://api.zaptec.com/oauth/token failed with status 503
ConfigEntryError: ... 503 Service Temporarily Unavailable
```

Reported by @sveinse in #392, who correctly suspected a race: setup issues a second `/oauth/token` POST within ~200 ms of the first, and the server occasionally answers that one with a transient `503`.

## Root cause

1. **No retry on transient 5xx for the token request.** `_refresh_token` branched only on `200` (return) and `400` (auth error); every other status — including `503` — fell straight through to `raise RequestError(...)` on the *first* response. The retry loop in `_request_worker` only re-looped on `TimeoutError` / `ClientConnectionError`, never on an HTTP status.
2. **A transient error mapped to a permanent setup failure.** In `async_setup_entry`, a `RequestError` (subclass of `ZaptecApiError`) became `ConfigEntryError`, which HA treats as *permanent* (no auto-retry). Only timeout/connection errors mapped to `ConfigEntryNotReady`. That's why only a restart recovered.

## The fix

**A — Retry transient statuses in `_request_worker`** (`zaptec/api.py`, `zaptec/const.py`)
- New `RETRYABLE_HTTP_STATUSES = frozenset({HTTPStatus.TOO_MANY_REQUESTS, HTTPStatus.BAD_GATEWAY, HTTPStatus.SERVICE_UNAVAILABLE, HTTPStatus.GATEWAY_TIMEOUT})`.
- The worker now retries those statuses for **all** methods using the existing exponential backoff, honoring a `Retry-After` header when present (as a floor on top of the computed backoff, never shortening it). On the final attempt it still yields the response so the caller raises its normal `RequestError(status)`. This fixes the token POST *and* every `request()` call from one place.
[...]

PR #391, opened 2026-07-08
Add offline unit tests for the Zaptec API client

rhammen:
## Summary

Adds **61 offline unit tests** for the vendored Zaptec API client (`custom_components/zaptec/zaptec/api.py`), taking statement coverage from ~13% to **~66%**. Until now the only api test (`test_api`) required a live login, so the core request/retry machinery and pure logic had effectively **no coverage** when running with `SKIP_ZAPTEC_API_TEST` / in CI.

The new tests use a small fake `aiohttp.ClientSession` and run without credentials or network access.

This is the first of a planned series to raise `api.py` to high-quality code (see the follow-up section). It's the safety net that makes the later type/correctness refactors reviewable. **No production code changes here — tests only.**

### What's covered
- **`request()` status handling** — 200 JSON, 204 → bytes, undecodable JSON → `RequestDataError`, error status → `RequestError` (with code), 500 GET retry-to-exhaustion vs 500 POST immediate raise, 401 → token refresh → retry
- **`_request_worker` retry/backoff** — connection errors and timeouts retried, then surfaced as `RequestConnectionError` / `RequestTimeoutError`
- **`login()` / `_refresh_token()`** — token stored and sent on later requests; 400 → `AuthenticationError`
- **`state_to_attrs`** — keydict mapping, `Value`/`ValueAsString` precedence, missing key/value skipping, `excludes`, duplicate-last-wins
- **`set_attributes`** — `ATTR_TYPES` conversion, snake_case keys, conversion-failure fallback, update-in-place
- **`is_command_valid`** — all resume/stop branches, plus a characterization test documenting the current `int(None)` behavior when `FinalStopActive` is absent (flagged for a later correctness fix)
- **`command()` / settings wrappers** — named id, numeric id, `authorize_charge` alias, unknown-command error; `set_settings`, `set_permanent_cable_lock`, `set_hmi_brightness` (URL + payload)
- **Installation current setters** — `set_limit_current` (availableCurrent, missing-arg, partial-phase, out-of-range) and `set_three_to_one_phase_switch_current`
- **`poll_info` / `poll_state`** — happy paths, `Charger.poll_info` 403 → charger-list fallback and non-403 re-raise, `Installation.poll_info` logo stripping
- **`stream_update` routing** — matching charger, unknown / missing / all-zero charger ids
- **mapping + lifecycle** — register/unregister/contains/qual_id, objects/installations/chargers/iter/len, `poll` dispatch + unknown-object error, async context manager

### Notes
[...]

Issue #162, opened 2025-02-03
Time is offset by 1 hour


theilgaard:
I believe reported charge in the time bucket is offset by 1 hour.

I recently installed a Pow K from amsleser.no to monitor my power usage, and can see the individual tracking from the Zaptec charger is arriving 1 hour later than the pulse reading from the amsleser Pow K. 

![Image](https://github.com/user-attachments/assets/b83bf60e-fbe5-4ee5-8cd6-35ee5af0aa9a)

Is there a way to offset this?

sveinse:
Unfortunately there is no way in HA to register data for the past, so a measurement happens when it happen. When Zaptec reports the hourly usage, it updates it too late and thus the data gets offset with one hour.

Its a common issue if you google it. https://community.home-assistant.io/t/energy-dashboard-is-shifted-one-hour/685596

theilgaard:
Ouch, so data is stored at the time of receiving, with no separate column/field in the messages for datetime? That seems like a massive blunder when working with time sensitive data.

sveinse:
It is, but it is a design decision made by HA. I recommend pursuing HA channels to get it changed.

In control engineering it is very common that when a data sample is made it takes some time before the sample is available and communicated, so I do agree that this is a weakness.

falver:
This is an issue of the Zaptec cloud platform combined with the lacking ability of HA to backdate data. The Zaptec platform is only updating a few values on a realtime basis to reduce the necessary processing power. All other values are "slow changing" and it might take minutes or up to an hour to process the data from the charger.

The solution (for now) is to use the "current charging session" entity. This sensor is updated realtime instead of on an hourly basis and in this way it maps the consumption to the correct time.

It does reset to 0 after every charging session but the energy dashboard can handle this correctly.

PR #396, opened 2026-07-12
Add per-charger energy statistics feed to fix Energy Dashboard hour-misattribution (fixes #300)

### LINKED ISSUE #300: Add state-less statistics entity for use with Energy Dashboard
Inspired by https://www.home-assistant.io/integrations/tibber/

This allows for properly backdating hourly power consumption which solves the offset issue seen in #162 and #103

Will require implementing the [chargehistory](https://docs.zaptec.com/reference/api_chargehistory_get)-api, and probably pagination as well.

rhammen:
## ⚠️ Stacked on #391, #393, #394, #395

This branch is based on my fork's `master`, which already includes #391, #393, #394, and #395 (all currently open, not yet merged upstream). Until those merge, this PR's diff will show all of their commits too. **Please merge #391 → #393 → #394 → #395 first** (in that order, as each stacks on the previous) — this PR's diff will then automatically shrink to just the 15 commits that are actually new here. I'm opening it now rather than waiting so it's visible/reviewable in parallel; happy to rebase once the others land if that's preferred.

## Summary

Fixes #300 by giving each tracked charger a **second, independent statistics feed** for the Energy Dashboard — sourced from Zaptec's own archived (completed) charge-session history via `/api/sessions/archived`, rather than from live polling of the charger's state.

The existing `Energy Meter` sensor's hourly statistics are built from whatever value the sensor happens to hold each time Home Assistant polls it — so consumption that happens near the end of an hour, or during a period of slow/delayed polling, gets attributed to the *next* hour instead. This is issue #300 (and the older #162): the Energy Dashboard shows real usage up to an hour later than it actually happened, and there's no way to fix that from the live sensor alone, since by the time HA polls it the information about *when within that hour* the energy was used is already lost.

Zaptec's cloud backend, however, already knows exactly when each session's energy was consumed — it exposes per-session hourly-ish meter checkpoints via `energyDetails` on `GET /api/sessions/archived`. This PR imports that data directly into Home Assistant's long-term statistics store as a new external statistic (`zaptec:energy_<charger id>`), which shows up in the Energy Dashboard's source picker as **"`<charger name>` Energy"**. It runs independently of the integration's existing live-state coordinators, polls hourly, and backfills up to 2 years of history on first run.

## Why an external statistic instead of fixing the existing sensor

The `Energy Meter` sensor's *live* value is still exactly what it should be — a real-time cumulative meter reading. The dashboard's hour-misattribution is purely a byproduct of *when HA happens to poll it*, not something a sensor definition change can fix in general. Home Assistant's own external statistics API exists precisely for this: a way to backfill *correctly-timed* historical statistics from an authoritative external source (the cloud's own session archive) without needing a live entity behind it at all — the same pattern integrations like Opower use for cost statistics. This also sidesteps the documented "Adjust sum" workaround needed today when using _Session total charge_ as a stand-in.

## What's implemented

- **`Zaptec.request()`** gained a `params` kwarg for query-string support (needed for the new endpoint's `From`/`To`/`Cursor`/`PageSize` params).
- **`Charger.get_archived_sessions()`** — cursor-paginated client for `/api/sessions/archived`, filtered by session *end* time (per the endpoint's actual semantics, confirmed against the official docs, not just `swagger.json`, which is missing/wrong on several details for this endpoint).
- **`validate.py`** gained a permissive (`extra="allow"`) pydantic model for the endpoint's response, since several of its fields' exact shapes aren't documented and shouldn't hard-fail the whole page if unconfirmed.
- **`bucket_sessions_hourly()`** (`statistics.py`) — the core pure function: turns a session's `energyDetails` into hourly `StatisticData` points, chaining onto a running sum. This went through substantial revision based on live testing against a real account (see below) - it's more subtle than it looks:
  - Each `energyDetails` point's `energy` field is the *incremental delta* since the previous point, not a cumulative running total (confirmed by cross-checking a session's `energyDetails` against its own OCMF-signed `sessionSignature` meter readings - independently signed data, not just an assumption).
  - Zaptec reports on a fixed hourly schedule while a session has metering activity, but silently skips ticks while paused/idle - so gaps between reports can span many hours (confirmed: real gaps of 4 and 17 hours in live data). A delta following such a gap is attributed to the hour *immediately before* the report that closed the gap (confirmed against the charger's own power sensor - the real energy consistently landed there, not spread across the gap and not at its start), clamped so a session's real (irregularly-timed) *end* isn't walked back past the previous report's hour.
  - A small tolerance window absorbs clock/reporting jitter of a few seconds right at an hour boundary, so it stays correct regardless of firmware/device timing precision - not just for the one device this was tested against.
  - Voided/aborted sessions are skipped (no meaningful energy, per the API docs). Sessions without `energyDetails` (pre-3.2 firmware) fall back to a single point at `endDateTime` using the session's total `energy`.
- **`ZaptecStatisticsCoordinator`** (`statistics.py`) — a `DataUpdateCoordinator` per charger, independent of the existing live-state coordinators so a failure here (e.g. a non-Owner account getting a 403, since this endpoint requires the Owner role) degrades only this one feed rather than the whole integration. Uses `async_refresh()` (not `async_config_entry_first_refresh()`) for the same reason. Resumes from `get_last_statistics()` on every poll rather than re-scanning all history.
- Wired into `manager.py`/`__init__.py`: one coordinator per tracked `Charger`, refreshed alongside the existing device coordinators.
- README updated to point users at the new "`<charger name>` Energy" statistic instead of the documented `Energy Meter`/`Session total charge` workaround.

## Testing

- Extensive unit tests for `bucket_sessions_hourly()` covering hour-bucketing, the resume/`after`-cutoff logic (including a regression test for a double-counting bug caught during review), voided/aborted skipping, the legacy fallback path, the multi-hour-gap attribution rule, and the session-end clamping edge case.
- Unit tests for `ZaptecStatisticsCoordinator` covering first-run backfill, resume, pagination, the 403/Owner-role soft-skip vs. hard-fail distinction, and HA-version compatibility for the `unit_class` metadata field (present in newer HA core, absent before).
- **Live-tested against a real account and charger** over several iterations, each surfacing and fixing a real issue that unit tests alone couldn't have caught:
  - The endpoint genuinely returns `camelCase` fields, unlike every other Zaptec endpoint's `PascalCase` (confirmed via a live `ValidationError`).
  - Older HA core versions (tested against 2025.10.2) don't have the `unit_class` statistics-metadata field at all; newer ones increasingly require it - handled with runtime feature detection instead of a hard version pin.
  - The `energyDetails`-is-a-delta and gap-attribution findings above, confirmed by diffing real archived-session data against both the session's own OCMF-signed meter readings and the charger's independent live power sensor.
  - Cross-checked the final result against an existing utility-meter-based tracker on the same charger and the raw power sensor as ground truth, for two full days of real usage, before considering this ready.

## Test plan for reviewers

- [ ] `SKIP_ZAPTEC_API_TEST=true pytest tests -q` passes (233 tests, plus 3 skipped needing live credentials)
- [ ] `ruff format --diff` and `ruff check` clean on all touched files
- [ ] After merging, verify a real charger's "`<charger name>` Energy" statistic appears in Settings → Dashboards → Energy → Add Consumption, and that a non-Owner account gets a clear warning log instead of a setup failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ Confirmed to work with Zaptec products
   [here](https://www.home-assistant.io/common-tasks/general/#defining-a-custom-polling-interval),
   will have unexpected effects. If the automatic polling is turned off, not all
   the data in the integration will update properly.
-* Using the _Energy Meter_ entity as an input to the Energy Dashboard will give values that are delayed by 1 hour
-  in the graphs (see [issue 162](https://github.com/custom-components/zaptec/issues/162) for details).
-  There is a plan to solve this in [issue 300](https://github.com/custom-components/zaptec/issues/300), but until that is implemented,
-  a workaround is to use the more frequently updated _Session total charge_ entity instead. This reduces the delay-issue,
-  but has a separate drawback where a restart of Home Assistant during a charging session can give a fake spike in the logged
-  consumption that needs to be manually edited using "Adjust sum" in the Statistics tab of the Developer tools dashboard.
+* Using the _Energy Meter_ entity directly as an input to the Energy Dashboard will still give values delayed by up to an hour
+  in the graphs, since that entity reflects live (polling-delayed) state (see [issue 162](https://github.com/custom-components/zaptec/issues/162)).
+  Each tracked charger now also gets a separate, invisible statistics feed (backdated hourly from Zaptec's charge history,
+  see [issue 300](https://github.com/custom-components/zaptec/issues/300)) that appears in the Energy Dashboard's device picker
+  as "<charger name> Energy" - use that entry instead of _Energy Meter_ or _Session total charge_ for accurate,
+  correctly-timed consumption graphs.
 
 
 # Installation and setup
--- a/custom_components/zaptec/__init__.py
+++ b/custom_components/zaptec/__init__.py
@@ -25,9 +25,11 @@ from .const import (
 from .coordinator import ZaptecUpdateCoordinator, ZaptecUpdateOptions
 from .manager import ZaptecConfigEntry, ZaptecManager
 from .services import async_setup_services, async_unload_services
+from .statistics import ZaptecStatisticsCoordinator
 from .zaptec import (
     RETRYABLE_HTTP_STATUSES,
     AuthenticationError,
+    Charger,
     Installation,
     RequestConnectionError,
     RequestError,
@@ -184,8 +186,20 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             ),
         )
 
+    # Setup the statistics coordinators, one per tracked charger, to backdate
+    # hourly energy consumption into HA's Energy Dashboard statistics (fixes
+    # the live sensor's misattributed-hour delay, see issue #300).
+    for deviceid in tracked_devices:
+        zaptec_obj = zaptec[deviceid]
+        if isinstance(zaptec_obj, Charger):
+            manager.statistics_coordinators[deviceid] = ZaptecStatisticsCoordinator(
+                hass, entry=entry, charger=zaptec_obj
+            )
+
     # Initialize the coordinators
     for co in manager.all_coordinators:
+        await co.async_config_entry_first_refresh()
+    for co in manager.statistics_coordinators.values():
         await co.async_config_entry_first_refresh()
 
     # Done setting up, change back to not log all updates. Having this enabled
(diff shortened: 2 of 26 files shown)


sveinse:
This PR is the oldest (lowest number) PR which is marked as ready. However the description state it depends on #391 #393, #394 and #395. The two first are merged, the two latter is in draft mode. This PR must thus either be paused as well or updated to not depend on these draft PRs. Please advise.

(This is generally why stacked PRs can become a little slow, as the merge sequence is fixed by the PR order.)

rhammen:
yes, I also noticed the issue.
I have been working on getting the changes to a new branch, independent of 394 and 395.
I propose to close this PR, and I will open a new one based on that new branch.
Agreed?

sveinse:
ok, please do. Just tag me on any PR that is ready for review. Thanks.

rhammen:
closing this PR, will create a new PR based on different branch.

Issue #311, opened 2025-08-29
Block calls to installation/update API if user does not have owner or service role


steinmn:
installation/update requires owner or service priviliges (https://docs.zaptec.com/reference/api_installation_id_update_post)

Also add explanation of this requirement to readme and error log/popup.

steinmn:
After trying out a test user that only had the User role, it seems like there isn't all that much to block, rather the question is if it is possible to do anything useful at all without at least owner privileges. The following polling calls were blocked:

- api.zaptec.com/api/installation/<--Inst[fa6acf]-->/hierarchy
- api.zaptec.com/api/chargerFirmware/installation/<--Inst[fa6acf]-->
- api.zaptec.com/api/chargers/<--Charger[2752a9]-->
- api.zaptec.com/api/chargers/<--Charger[2752a9]-->/state

With neither the service bus nor the state api available, there really isn't much useful info left.

thecoldwine:
@steinmn I have a question — if let's say we had an API with notifications for the installation (not service bus, but something else), what info would you like to see there? I am curious, because most likely you don't need every observation or change on the installation.

thecoldwine:
@sveinse I will communicate this to product team, thanks a lot.

sveinse:
@thecoldwine any updates from your side on this issue?

PR #397, opened 2026-07-12
Relax over-strict validation models to fields api.py actually uses (fixes #359)

### LINKED ISSUE #359: Review all validation models and make sure we don't require something we don't use.
Follow-up to #351, #354 and #357

Also check if there is anything we currently don't require in validation that should be added because the lack of it would cause other issues.

Make sure to run a test with a User-role-only test-user before merging anything. We may also want to consider distinguishing between FullInstallation (returned if owner role) and LimitedInstallation (returned if only user role), since the Installations list will include all installations the user has access to, even if they do not use it in our integration (because they only have the user role for it).

### LINKED ISSUE #392: Setting up Zaptec integration fails
When setting up the integration in a fresh system, such as a dev-container, and adding a Zaptec account returns the error below. This is running current master (as of 2026-07-08). What's interesting is that restarting HA will successfully start Zaptec. Without looking too much at the data yet, I think this might be a race.

```
2026-07-08 21:09:02.267 DEBUG (MainThread) [custom_components.zaptec.zaptec.api] @@@  REFRESH TOKEN
2026-07-08 21:09:02.267 DEBUG (MainThread) [custom_components.zaptec.zaptec.api] @@@  REQUEST POST to 'https://api.zaptec.com/oauth/token' data length 3
2026-07-08 21:09:02.479 DEBUG (MainThread) [custom_components.zaptec.zaptec.api] @@@  RESPONSE 200 length 1408
2026-07-08 21:09:02.480 DEBUG (MainThread) [custom_components.zaptec.zaptec.api]      TOKEN OK
2026-07-08 21:09:02.481 DEBUG (MainThread) [custom_components.zaptec] Setting up entry 01KX1S03WG2B91J19FXDXDJG4D: {'username': '********', 'password': '********'}
2026-07-08 21:09:02.481 DEBUG (MainThread) [custom_components.zaptec.zaptec.api] @@@  REFRESH TOKEN
2026-07-08 21:09:02.481 DEBUG (MainThread) [custom_components.zaptec.zaptec.api] @@@  REQUEST POST to 'https://api.zaptec.com/oauth/token' data length 3
2026-07-08 21:09:02.515 DEBUG (MainThread) [custom_components.zaptec.zaptec.api] @@@  RESPONSE 503 length 197
2026-07-08 21:09:02.515 ERROR (MainThread) [custom_components.zaptec] Zaptec API error: POST request to https://api.zaptec.com/oauth/token failed with status 503: <ClientResponse(https://api.zaptec.com/oauth/token) [503 Service Temporarily Unavailable]>
<CIMultiDictProxy('Server': 'nginx/1.29.8', 'Date': 'Wed, 08 Jul 2026 21:09:02 GMT', 'Content-Type': 'text/html', 'Content-Length': '197', 'Connection': 'keep-alive')>

2026-07-08 21:09:02.516 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Zaptec for zaptec: POST request to https://api.zaptec.com/oauth/token failed with status 503: <ClientResponse(https://api.zaptec.com/oauth/token) [503 Service Temporarily Unavailable]>
<CIMultiDictProxy('Server': 'nginx/1.29.8', 'Date': 'Wed, 08 Jul 2026 21:09:02 GMT', 'Content-Type': 'text/html', 'Content-Length': '197', 'Connection': 'keep-alive')>
  File "/workspaces/zaptec/custom_components/zaptec/__init__.py", line 82, in async_setup_entry
    await zaptec.login()
  File "/workspaces/zaptec/custom_components/zaptec/zaptec/api.py", line 1042, in login
    await self._refresh_token()
  File "/workspaces/zaptec/custom_components/zaptec/zaptec/api.py", line 1089, in _refresh_token
    raise log_exc(
    ...<4 lines>...
    )
custom_components.zaptec.zaptec.exceptions.RequestError: POST request to https://api.zaptec.com/oauth/token failed with status 503: <ClientResponse(https://api.zaptec.com/oauth/token) [503 Service Temporarily Unavailable]>
<CIMultiDictProxy('Server': 'nginx/1.29.8', 'Date': 'Wed, 08 Jul 2026 21:09:02 GMT', 'Content-Type': 'text/html', 'Content-Length': '197', 'Connection': 'keep-alive')>


The above exception was the direct cause of the following exception:

  File "/home/vscode/.local/lib/python3.14/site-packages/homeassistant/config_entries.py", line 769, in __async_setup_with_context
    result = await component.async_setup_entry(hass, self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/zaptec/custom_components/zaptec/__init__.py", line 91, in async_setup_entry
    raise ConfigEntryError from err
homeassistant.exceptions.ConfigEntryError: POST request to https://api.zaptec.com/oauth/token failed with status 503: <ClientResponse(https://api.zaptec.com/oauth/token) [503 Service Temporarily Unavailable]>
<CIMultiDictProxy('Server': 'nginx/1.29.8', 'Date': 'Wed, 08 Jul 2026 21:09:02 GMT', 'Content-Type': 'text/html', 'Content-Length': '197', 'Connection': 'keep-alive')>
```

rhammen:
## ⚠️ Stacked on #391, #393, #394, #395 — please read before merging

This branch is built on top of this fork's `master`, which already includes #391 ("Add offline unit tests for the Zaptec API client"), #393 ("Retry transient server errors on setup", fixes #392), #394 ("coordinator/entity test coverage"), and #395 ("platform entity test coverage") — all still open against this repo at the time of writing.

**Recommended merge order:** #391 → #393 → #394 → #395 → this PR. Until those four land, this PR's diff/commit list will show their commits too — that's expected for a stacked PR, not a duplicate. Once they're merged, this PR's diff will shrink to just its own 6 commits.

## Summary

Fixes #359. Continues the work from #354/#357 (fixed in #355): several pydantic models in `validate.py` required API response fields that `api.py` never actually reads, which turns a harmless reduced or unusual response into a hard integration failure instead of degrading gracefully.

**Governing rule applied to every model in this PR:** a field is required only if `api.py` reads it via a hard subscript (`d["Key"]`, no `.get()`/default). Everything else is relaxed to optional. This was applied by reading every hard subscript in `api.py` against each response shape, not by guessing.

## What this fixes

- **`Circuit.MaxCurrent`** (new, required) — `api.py` hard-indexes `circuit["MaxCurrent"]` in `Installation.build()`, but this field was never declared on the `Circuit` model at all. A hierarchy response missing it would previously pass validation and then crash with a raw, unlogged `KeyError` deep in `build()`. This is the clearest concrete instance of the bug class #359 asks to close.
- **`HierarchyCharger`** (new model) — the charger stubs embedded in a hierarchy response's `Circuit.Chargers` were being validated against the same `Charger` model used for the full `/chargers` endpoint, requiring `Name`/`Active`/`DeviceType` that aren't guaranteed there. Split into a minimal model requiring only `Id`+`DeviceType` (see below for why `DeviceType` stayed required).
- **`Installation`/`Installations`** — only `Id` stays required; `Active`, `CurrentUserRoles`, `InstallationType`, `NetworkType`, `Pages` relaxed (none are hard-read by `api.py`).
- **`Charger`** (`/chargers`) — `Id`/`DeviceType` stay required, `Name`/`Active` relaxed.
- **`Circuit`/`Hierarchy`** — `Name`/`Chargers` (Circuit) and `Id`/`Name`/`NetworkType` (Hierarchy) relaxed; `api.py`'s `Installation.build()` hardened to match (`.get()` instead of hard subscripts for the now-nullable fields).
- **`ChargerFirmware`** — only `ChargerId` stays required. `poll_firmware_info()` already has defensive code (`fm.get("CurrentVersion") is None` etc.) for chargers added but not yet initialized — but validation ran *before* that code and rejected the response outright, so the defensive branch could never actually execute. Now it can.

**One bug caught only by the final whole-branch review, not any single task:** `HierarchyCharger` initially dropped `DeviceType` too, reasoning that only `Id` is read from it at parse time. That missed that `api.py`'s standalone-charger merge loop skips re-merging any charger already found via the hierarchy (`if chgid in installation_chargers: continue`), so a hierarchy-sourced charger's data comes *only* from its hierarchy stub — never from the fuller `/chargers` list. `Zaptec.build()` later hard-subscripts `chg["DeviceType"]` on every registered charger, so this would have turned a clean validation error into a raw `KeyError` crash — the exact anti-pattern this PR exists to close, just one level removed. Fixed before merging into this branch; see commit history for the full trace.

## What this does NOT do (explicitly out of scope)

- **No role-based blocking of API calls** — that's #311, a separate, larger behavioral change (proactively refusing to call `installation/update` etc. for non-Owner/Service accounts with a clear error, rather than just tolerating whatever comes back). This PR only fixes validation strictness.
- **No `FullInstallation`/`LimitedInstallation` model split** — raised as an idea in #359's issue body, deliberately not done: `api.py` never branches its logic on which fields are present, so a second model class would be an unused abstraction. Making the role-dependent fields optional on one model achieves the same practical result.
- **`InstallationConnectionDetails` (`messagingConnectionDetails`) and `ChargerState`/`CHARGER_STATES` are untouched** — the former is already marked deprecated with all its required fields hard-indexed and no bug reports; the latter was already correctly relaxed (only `StateId` required) before this PR.

## Live testing

Tested end-to-end against a real installation/charger across **four different Zaptec account role configurations** (separate HA config entries, debug logging enabled), not just unit tests:

| Roles (`CurrentUserRoles`) | `installation` | `hierarchy` | `chargerFirmware` | charger detail/state |
|---|---|---|---|---|
| `Owner, User, Maintainer` | full | full | full | full |
| `User` only | reduced (no `AuthenticationType`) | `204` (blocked) | `403` | `403` |
| `Maintainer` (the portal's "Service" role) only | full | full | full | full |
| `Owner` only | full | full | full | full |

No `ValidationError`, no crash, clean setup in all four cases.

**Honest caveat:** only the plain `User` role showed any reduction at all, and mostly via outright endpoint blocks (`403`/`204`) rather than the "200 with fields silently dropped" pattern this PR's relaxations target — the one confirmed instance of that pattern is `Installation.AuthenticationType` (the original #357 symptom, already fixed pre-this-PR in #355). So while every change here is justified by real evidence (the #354/#357 bug reports, the Zaptec API docs, and full-coverage reading of `api.py`'s hard subscripts), most of the specific relaxations in this PR — `Circuit.MaxCurrent`, `HierarchyCharger.DeviceType`, the `ChargerFirmware` fields, and most of the `Installation`/`Charger` field relaxations — were not freshly proven against a live reduced-but-200 response in this round of testing, because the accounts available for testing didn't happen to trigger that specific pattern on those endpoints. They remain correct by construction; a live account that does trigger it would be needed to close that gap fully.

## Test plan

- [x] `SKIP_ZAPTEC_API_TEST=true pytest tests -q` — 214 passed, 2 skipped, 22 errors (pre-existing, documented DNS-fixture gap in `test_zconst.py`/`test_redact.py`, unrelated to this change)
- [x] `ruff format --diff` / `ruff check` clean on all touched files
- [x] Each task individually reviewed (spec compliance + code quality) plus a final whole-branch review; one cross-task bug found and fixed before merge (see above)
- [x] Live end-to-end testing against a real installation across 4 account role configurations (see table above)


custom_components/zaptec/zaptec/validate.py
```diff
@@ -16,27 +16,43 @@ class Installation(BaseModel):
...
+    via the installation hierarchy, so a hierarchy-sourced charger's
+    attributes come only from this stub -- including DeviceType, which is
+    later hard-subscripted for every registered charger.
+    """
 
     model_config = ConfigDict(extra="allow")
     Id: str
-    Name: str
-    Active: bool
     DeviceType: int
```
steinmn, line 56: Given that Name and Active are now optional (can be None), adding HierarchyCharger adds unnecessary complexity.

custom_components/zaptec/zaptec/validate.py
```diff
@@ -61,17 +77,20 @@ class Circuit(BaseModel):
...
 
 
 class Hierarchy(BaseModel):
     """Pydantic model for the hierarchy of Zaptec objects in an installation."""
 
     model_config = ConfigDict(extra="allow")
-    Id: str
-    Name: str
-    NetworkType: int
+    Id: str | None = None
```
steinmn, line 91: Id cannot be None, as far as I can tell

custom_components/zaptec/zaptec/validate.py
rhammen: Addressed in #415 which supersedes this PR

custom_components/zaptec/zaptec/validate.py
rhammen: Addressed in #415 which supersedes this PR

rhammen:
Closing in favour of #415, which un-stacks this work onto `master` (no test-PR ancestry), addresses the review remarks above, and also fixes #400.

PR #398, opened 2026-07-12
fix: force OS DNS resolver in zaptec_constants test fixture

rhammen:
## Summary
- `aiohttp.ClientSession()` auto-switches to `aiodns`/`pycares`'s `AsyncResolver` whenever those packages are importable, bypassing the OS resolver stack
- In some dev environments (e.g. native Windows without a working async DNS path) that resolver can't reach DNS (`Could not contact DNS servers`), even though OS-backed lookups (`urllib`, `ThreadedResolver`) work fine from the same shell
- This makes the `zaptec_constants` fixture fail regardless of `SKIP_ZAPTEC_API_TEST`, taking down `tests/zaptec/test_zconst.py` and `tests/zaptec/test_redact.py`
- Fix: force `aiohttp.resolver.ThreadedResolver()` via an explicit `TCPConnector` for the fixture's `Zaptec` client — no behavior change when `aiodns` isn't installed, and no change to what's actually tested

## Test plan
- [x] `pytest tests/zaptec/test_zconst.py tests/zaptec/test_redact.py -q` — 23 passed (previously errored in the affected environment)
- [x] `pytest tests -q` — 231 passed, 2 skipped
- [x] `ruff format --diff` / `ruff check` clean on `tests/conftest.py`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@
 import asyncio
 import os
 
+import aiohttp
 import pytest
 
 from custom_components.zaptec.zaptec.api import Zaptec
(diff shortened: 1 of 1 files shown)


steinmn:
Why run natively on Windows and not in the dev-container that has been tailored to make sure all developers are using the same dev-environment? See [DEVELOPMENT.md#setting-up-the-dev-container](https://github.com/custom-components/zaptec/blob/master/DEVELOPMENT.md#setting-up-the-dev-container)

rhammen (commented):
Closing this — you're right, and I've settled on the devcontainer as the answer rather than accommodating native Windows in tracked files.

Closing without a replacement; native-Windows local dev isn't a target this repo needs to support in tracked files.

Issue #274, opened 2025-08-08
Error while setting up subset of chargers


sveinse:
If selecting a subset of chargers the setup will fail with errors. This is due to missing device coordinators for any chargers that is _not_ selected, while the setup loop is iterating over all chargers present.

```
2025-08-08 10:23:09.372 ERROR (MainThread) [homeassistant.components.binary_sensor] Error while setting up zaptec platform for binary_sensor: 'xxx'
Traceback (most recent call last):
  File "/home/vscode/.local/lib/python3.13/site-packages/homeassistant/helpers/entity_platform.py", line 382, in _async_setup_platform
    await asyncio.shield(awaitable)
  File "/workspaces/zaptec/custom_components/zaptec/binary_sensor.py", line 108, in async_setup_entry
    entities = entry.runtime_data.create_entities_from_zaptec(
        INSTALLATION_ENTITIES,
        CHARGER_ENTITIES,
    )
  File "/workspaces/zaptec/custom_components/zaptec/__init__.py", line 470, in create_entities_from_zaptec
    self.create_entities_from_descriptions(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        installation_descriptions,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
        obj,
        ^^^^
        info,
        ^^^^^
    )
    ^
  File "/workspaces/zaptec/custom_components/zaptec/__init__.py", line 415, in create_entities_from_descriptions
    coordinator=self.device_coordinators[zaptec_obj.id],
                ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'xxx'
```

PR #399, opened 2026-07-14
Remove stale devices for deselected chargers after reconfigure (fixes #272)

### LINKED ISSUE #272: Charger present after reconfiguration and deselected a charger
When using reconfiguring the integration and a smaller set of chargers are selected, any old chargers that are no longer configured are not removed. It should be considered if there is need for a cleanup after reconfiguration. -- Which I think is doing a full reload, so we should probably check that all devices (chargers) that belongs to Zaptec are the user configured chargers on startup.

How to repro:

1. If you have only one charger, a small modification is needed to fake a second charger:

    ```py
    # Add to L101 of config_flow.py
    result[0]["foobar"] = "Foobar charger"
     ```

2. Run HA and select reconfigure in Zaptec, and check "manual_select". Submit
3. Select "Foobar charger" and submit
4. Observe that the device is still present

> [!NOTE]
> Remember to reconfigure it back to a real charger before removing the foobar code, otherwise the configuration will not work properly.

rhammen:
## Summary
- Fixes #272 - a charger deselected via reconfigure (manual_select) kept its device and entities in the registry forever instead of being cleaned up.
- The existing device-registry cleanup pass in `async_setup_entry` (previously inline, now extracted into `_cleanup_stale_devices`) only removed devices with zero entities or devices matching a deprecated "Circuit" device id. It never checked whether a device's zaptec id was still tracked after a reconfigure. This adds that check.
- Depends on #274 (fixed by #275, already merged), which made entity creation skip untracked zaptec objects - that's why a deselected charger gets no new entities on reload, but its old entities were never explicitly removed.
- **Safety guard:** the new removal only runs once `ZaptecManager.first_time_setup` confirms every selected charger was present in the Zaptec API response this session, and only in manual-select mode. This is deliberate: in track-all mode there's no concept of "deselection" to compare against, so a transient/partial API response (outage, momentary blip) could otherwise be mistaken for the user removing a charger and permanently delete a still-selected charger's device and entities. Two review passes on this branch specifically probed for that failure mode before landing.

## Dependency note

This branch is based on our fork's `master`, which already includes commits from these open upstream PRs: #391, #393, #394, #395, #398. Please merge those first (in that order) - until then this PR's diff will include their changes too.

## Test plan
- [x] `SKIP_ZAPTEC_API_TEST=true pytest tests -q` - 241 passed, 2 skipped
- [x] `ruff format custom_components tests --diff` - clean
- [x] New/updated tests cover: no-entities removal, deprecated-circuit removal, tracked-device-kept, deselected-charger removal (#272's scenario), skip-removal when the API response is incomplete, tracked-installation-kept, and the full manual-select-vs-track-all decision matrix


rhammen:
Superseded by #424.

This PR was written against the test architecture that #414 replaces, so rebasing it would be more churn than value. #424 is the same fix ported onto #414's pytest-homeassistant-custom-component harness, and additionally fixes a domain-guard bug in the identifier loop that the port surfaced.

Closing in favour of #424.

PR #401, opened 2026-07-14
Block Owner/Service-only writes and warn on insufficient role (fixes #311)

### LINKED ISSUE #311: Block calls to installation/update API if user does not have owner or service role

rhammen:
> ⚠️ **Stacked PR — please read before merging.** This branch is built on top of several other open PRs against this fork's `master`, in order: #391 → #393 → #394 → #395 → #397. Because of that, the diff below includes all of their changes too, not just this PR's own work — **the actual new content of this PR is only the last 4 commits** (`c8c6a6b`..`4560f10`): `custom_components/zaptec/zaptec/exceptions.py`, `__init__.py`, `api.py`, `custom_components/zaptec/coordinator.py`, `translations/en.json`, `README.md`, and the two new test files. Recommended merge order: #391 → #393 → #394 → #395 → #397 → this PR. Happy to rebase onto `master` directly once the earlier PRs land, if that's easier to review.

## Summary

Fixes #311. `installation/update` requires the Owner or Service role — but so do two other write endpoints this integration calls that weren't mentioned in the issue. I checked `docs.zaptec.com/reference/*.md` for every write endpoint in `api.py` and found:

| Endpoint | api.py method(s) | Docs say |
|---|---|---|
| `installation/update` | `Installation.set_limit_current`, `set_three_to_one_phase_switch_current` | "requires owner or service permissions" |
| `chargers/{id}/update` | `Charger.set_settings` | "requires owner or service permissions" (identical wording) |
| `chargers/{id}/SendCommand/{id}` | `Charger.command` (restart, stop, resume, deauthorize, upgrade_firmware) | "owner or service-level access" |
| `chargers/{id}/authorizecharge` | `Charger.authorize_charge` | not documented anywhere — `api.py` already comments it `# NOTE: Undocumented API call` |
| `chargers/{id}/localSettings` | `Charger.set_hmi_brightness`, `set_permanent_cable_lock` | not documented anywhere |

So this PR gates all three documented endpoints, and deliberately leaves the two undocumented ones alone — there's no evidence for what role (if any) they require, and guessing wrong risks blocking something that actually works.

## What this does

1. **Blocks the write, with a clear error, before the API call is made.** A new `ZaptecBase._require_write_role()` helper (shared by `Installation` and `Charger`, since both carry `CurrentUserRoles`) checks the account's role and raises `InsufficientRoleError` naming the exact action and role needed, e.g. `"Setting the installation current limit requires the Owner or Service role on Installation[abc123] (current role: User)."`. This already flows through the existing generic exception handling in `number.py`, `services.py`, and `button.py` into a `HomeAssistantError` the user actually sees — no changes needed there.
2. **Proactively warns before anyone even tries.** The per-installation update coordinator now creates a Home Assistant [Repair](https://www.home-assistant.io/integrations/repairs/) issue (Settings → Repairs) when it detects insufficient role on that installation, naming the installation and the role it needs, with a link to the Zaptec Portal. It's non-fixable (informational) and can be dismissed with "Ignore" — dismissal survives across polls as long as role doesn't change (verified against Home Assistant's own `issue_registry` source: repeat `async_create_issue()` calls for the same issue don't reset `dismissed_version`).
3. **README** documents both behaviors.

## What this does NOT do

- **Does not refuse setup.** A `User`-only account still sets up and works for everything that doesn't need elevated access — this only affects the specific write calls above.
- **Does not gate `authorize_charge` or HMI brightness/cable lock** (`localSettings`) — undocumented endpoints, no evidence either way.
- **Repair issue is Installation-scoped only, not per-charger.** Role has been confirmed to vary *per charger* under one login (see #357: "I have one as Owner, and one as User") — but that's evidence of deliberate per-charger access management, not user confusion, so proactively warning per-charger felt like it'd mostly nag people who already know what they configured. The reactive error (point 1) already covers that case when someone actually tries a restricted charger's controls.
- **Does not change entity visibility.** Every existing entity is still created exactly as before; only actually invoking a gated write call behaves differently for insufficient-role accounts.

## Test plan

- [x] 16 new offline unit tests in `tests/zaptec/test_api.py` (`FakeSession`-based, no network) covering block/allow for all four gated methods across `User`/`Owner`/`Maintainer`/unknown roles, plus explicit negative tests proving `authorize_charge` and `set_hmi_brightness` stay ungated even for a `User`-only account.
- [x] 4 new tests in `tests/test_coordinator.py` covering Repair issue creation, clearing, the "never delete-then-recreate while role stays insufficient" dismissal-preservation contract, and that non-Installation coordinators never run the check.
- [x] Full suite: 234 passed, 2 skipped, 22 errors (pre-existing DNS-resolution gap in this dev environment, unrelated to this change — see `CLAUDE.md`).
- [x] `ruff format --diff` and `ruff check` clean on all changed files.


rhammen:
Superseded by #426, which ports this onto #414's test conventions and adds a per-charger warning.

PR #402, opened 2026-07-14
Refactor services.py: register once in async_setup, fix multi-account resolution

### LINKED ISSUE #253: Refactor services.py
`async_setup_services` should be run in `async_setup`, not in `async_setup_entry` (https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/action-setup/) and `async_unload_services` seems to have been dropped entirely (https://github.com/home-assistant/core/pull/146048#discussion_r2121194482). Haven't been able to find where the recommendation this comment refers to is, but I do see that `async_unload_services is no longer used anywhere in the core-repo.

Looking at some of the integrations in core that implement `async_setup_services`, they rarely contain more than `hass.services.async_register`-calls within the function itself.

Prerequisite:
Ensure good test coverage for services before refactoring to avoid introducing regressions.

Main tasks:

- Move call to `async_setup_services` from `async_setup_entry` to `async_setup`
- Move function definitions out of `async_setup_services`
- Investigate if `iter_objects` can be simplified and/or makes more sense as a function inside `ZaptecManager`
- remove `async_unload_services`

rhammen:
## Summary
- Fixes #253 - moves `async_setup_services` registration from `async_setup_entry` (once per config entry) to a new `async_setup` (once at HA startup), per HA's [action-setup quality-scale rule](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/action-setup/). `async_unload_services` is removed - nothing else called it, and services now live for HA's lifetime, not any one config entry's.
- Moves `iter_objects` and all eight `service_handle_*` coroutines out of the closure nested inside `async_setup_services` to module level in `services.py`, resolving the "too complex" `# noqa: C901` that pointed at this issue.
- **Also fixes a real bug this refactor uncovered**: `iter_objects` used to search only the one `ZaptecManager` captured in a closure at registration time. Because registration was guarded by `hass.services.has_service`, a user with two Zaptec accounts (two config entries) could never reach the second account's chargers via `zaptec.*` services - only whichever entry set up first ever got its manager wired in. `iter_objects` now resolves `service_call.hass` and searches every currently loaded config entry's manager via `hass.config_entries.async_entries(DOMAIN)` + `entry.runtime_data`.
- Investigated whether `iter_objects` belongs on `ZaptecManager` (per the issue's ask): decided against it - once services register globally, `iter_objects` has to search *across* every loaded entry's manager, so it can't be a method on any single `ZaptecManager` instance without that instance reaching outside itself to its siblings, which nothing else in the codebase does.

## Dependency note

This branch is based on our fork's `master`, which already includes commits from these open upstream PRs: #391, #393, #394, #395. Please merge those first - until then this PR's diff will include their changes too.

## Test plan
- [x] `SKIP_ZAPTEC_API_TEST=true pytest tests -q` - 233 passed, 2 skipped
- [x] `ruff format custom_components tests --diff` - clean
- [x] `ruff check --exclude custom_components/zaptec/zaptec/api.py` - clean (the CI-gating strict check)
- [x] New/updated tests cover: multi-account resolution (a charger reachable only through a second config entry's manager), a config entry with no `runtime_data` (not currently loaded) being skipped rather than crashing, plus all pre-existing service-handler and `iter_objects` error-path coverage carried over unchanged


rhammen:
Replaced by #419 , because test framework changed significantly (PR394 was replaced by PR414)

PR #403, opened 2026-07-23
fix: make pytest runnable natively on Windows dev environments

rhammen:
## Summary
- `pytest-homeassistant-custom-component` autoloads via a `pytest11` entry point on every pytest run once installed, and unconditionally imports `fcntl` - which doesn't exist on Windows, so pytest collection failed before a single test could run on native Windows dev setups.
- Adds `-p no:homeassistant` to `pyproject.toml` to block that autoload, and a new root `conftest.py` that re-loads the plugin explicitly, first installing Windows-only shims for `fcntl`, `resource`, and a `socket.socketpair` patch (needed because HA's async test harness relies on all three). pytest only honors `pytest_plugins` in the rootdir conftest, so this can't live in `tests/conftest.py`. All shims are guarded by `sys.platform == "win32"` and are no-ops on Linux/CI.
- With the plugin now actually loading, its global `pytest-socket` block also started blocking the `zaptec_constants` fixture's live HTTPS call to `api.zaptec.com` (`tests/conftest.py`) - scoped socket access back on for just that fixture's duration.
- That fixture's `asyncio.run()` call was also clearing the event-loop-policy's current loop on exit, which broke every subsequent test under HA's custom event loop policy (it raises instead of lazily recreating a loop like vanilla asyncio). Switched to an explicit `loop.run_until_complete()` that doesn't touch global event-loop-policy state.

None of this changes runtime behavior of the integration - it's dev/test tooling only, and every shim is inert on the Linux CI environment this repo already tests on.

## Test plan
- [x] `SKIP_ZAPTEC_API_TEST=true pytest tests -q` on Windows - previously failed to even collect (`ModuleNotFoundError: No module named 'fcntl'`), now 231 passed, 2 skipped
- [x] `ruff format custom_components tests conftest.py --diff` - clean
- [x] `ruff check --exclude custom_components/zaptec/zaptec/api.py` - clean (the CI-gating strict check)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,57 @@
+"""Repo-root conftest: loads pytest-homeassistant-custom-component explicitly.
+
+The plugin autoloads via a pytest11 entry point on every pytest run once
+installed, but importing it on Windows fails immediately (imports `fcntl`,
+Unix-only) before any test collects. `-p no:homeassistant` in pyproject.toml
+blocks that autoload; this file loads the plugin back explicitly, with
+Windows compatibility shims applied first. pytest only honors
+`pytest_plugins` in the rootdir conftest, so this cannot live in
+tests/conftest.py.
+"""
+
+import sys
+import types
+from typing import Any
+
+if sys.platform == "win32":
+    import pytest_socket
+
+    if "fcntl" not in sys.modules:
+        fake_fcntl = types.ModuleType("fcntl")
+        fake_fcntl.LOCK_SH = 1
+        fake_fcntl.LOCK_EX = 2
+        fake_fcntl.LOCK_NB = 4
+        fake_fcntl.LOCK_UN = 8
+        fake_fcntl.flock = lambda *_args, **_kwargs: None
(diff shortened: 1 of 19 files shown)


rhammen:
Closing - this was based on a wrong diagnosis. `pytest-homeassistant-custom-component` is not actually a dependency of this project (`requirements_test.txt` only lists `pytest`, `pytest-asyncio`, `pytest-mock`, `pytest-cov`), and this repo's own `tests/conftest.py` already uses hand-rolled `hass`/`FakeConfigEntry` mocks specifically to avoid needing that plugin's Windows-incompatible harness.

The plugin was only autoloading in my local dev environment because it happens to also be installed there for an unrelated sibling project - not something CI or other contributors would hit. As written, the fix doesn't help CI (which never had this problem) and actively breaks it, since the new root `conftest.py` unconditionally tries to import a package that genuinely isn't installed on CI.

If we later migrate this repo's tests to actually use the real HA test harness (which would be a legitimate improvement over the hand-rolled mocks), a Windows-compatibility shim like this would make sense as part of that change. As a standalone PR it doesn't belong here.

PR #404, opened 2026-07-23
i18n: add German (de) translation

rhammen:
## Summary
- Adds `custom_components/zaptec/translations/de.json`, a full German translation covering every key currently in `en.json` (config flow + all entity platforms).

## Note
AI-assisted translation, reviewed by me (a non-native German speaker) and it reads well. Still worth a native-speaker pass before merge to catch anything a non-native reviewer wouldn't.

## Test plan
- [x] Valid JSON
- [x] Key set matches `en.json` exactly (no missing or extra keys)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -167,6 +167,7 @@ Or by the snippet:
 
 ```py
 import json
+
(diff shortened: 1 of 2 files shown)


steinmn (approved):
lgtm

PR #405, opened 2026-07-23
i18n: add French (fr) translation

rhammen:
## Summary
- Adds `custom_components/zaptec/translations/fr.json`, a full French translation covering every key currently in `en.json` (config flow + all entity platforms).

## Note
AI-assisted translation, reviewed by me (a non-native French speaker) and it reads well. Still worth a native-speaker pass before merge to catch anything a non-native reviewer wouldn't.

## Test plan
- [x] Valid JSON
- [x] Key set matches `en.json` exactly (no missing or extra keys)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -167,6 +167,7 @@ Or by the snippet:
 
 ```py
 import json
+
 async with Zaptec(username, password) as zaptec:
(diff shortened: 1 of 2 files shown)


sveinse:
Since we don't have French in our integration yet, I'm inclined to just get it merged in and then we can rather deal with any issues if someone complains.

Issue #406, opened 2026-07-23
Ruff CI is fragile to version drift: select=["ALL"] + no ruff version pin will silently break every PR


rhammen:
## What's happening

`.ruff.toml` uses `select = ["ALL"]`, and neither `pyproject.toml` (`required-version = ">=0.15.0"`, no upper bound) nor `.github/workflows/validate.yaml` pin an exact ruff version - the workflow's `ruff format`/`ruff check` steps just fetch whatever the constraint allows, which is currently ruff 0.16.0.

This makes CI fragile in two compounding ways whenever ruff cuts a new release:

1. **New formatting behavior**: ruff 0.16.0 changed how it formats Python code blocks embedded in Markdown. Under 0.16.0, `ruff format --diff .` on current `master` reports `DEVELOPMENT.md` would be reformatted (a blank line inserted after `import json` in one code fence). This fails the first ("Run ruff format") step of the `Ruff validation` job.

2. **New rule auto-enabled by `ALL`**: ruff 0.16.0 also added `CPY001` (missing-copyright-notice). Because `select = ["ALL"]` opts into every rule ruff ships, this rule is now active repo-wide and isn't in `.ruff.toml`'s `ignore` list. Running `ruff check --exclude custom_components/zaptec/zaptec/api.py` (the gating check step) with 0.16.0 reports **39 errors**, one per file lacking a copyright header - across nearly the whole `custom_components/zaptec` tree and several test files.

CI hasn't visibly hit problem 2 yet only because problem 1 fails first and aborts the job before the `ruff check` steps run (confirmed: the failing job's log shows only the `Run ruff format` step executed). The moment `DEVELOPMENT.md` gets reformatted to satisfy step 1, the gating `ruff check --exclude api.py` step will immediately start failing on all 39 files from problem 2 - breaking every currently-open PR's CI, not just whichever PR happens to fix the Markdown formatting.

## How I found this

Investigating why PR #404 and #405 (translation additions, unrelated Python changes) were failing "Ruff validation" in CI despite passing locally with ruff 0.15.9 (the version in my local dev environment). Reproduced locally by installing ruff 0.16.0 in an isolated location and running the same two commands CI runs, against plain `master`.

## Suggested fixes (pick one, not prescribing an approach)

[...]

PR #407, opened 2026-07-24
Use requirements.txt for ruff version in ci-pipeline

steinmn:
Quickfix for #406 by using requirements.txt to define ruff version in ci-pipeline instead of pyproject.toml.

Also remove workaround for api.py, since there are no errors left.


sveinse (approved):
Approved. Thank you @steinmn for a very swift response.

PR #409, opened 2026-07-24
Add readme-section on undocumented API calls

### LINKED ISSUE #153: Support for enable/disable Zaptec Eco Mode
Zaptec seems to offer a charge optimization service called EcoMode that can be enabled via API or via the Zaptec portal, which can be useful for some users.

To override the EcoMode one needs to disable it, it would be very useful to be able to do this via HomeAssistent.

To enable/disable, it seems one uses the API calls:

curl 'https://api.zaptec.com/api/installation/000000000-1111-2222-333333333333333' -X PUT

with data like:
  "EnabledFeatures": 12,
  "Feature_PowerManagement_EcoMode_DepartureTime": 360,
  "Feature_PowerManagement_EcoMode_MinEnergy": 10,
  "Feature_PowerManagement_EcoMode_DeliveryArea": 8,

It would be super to have this feature in the integration.

steinmn:
Closes #153 (for now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ Confirmed to work with Zaptec products
   but has a separate drawback where a restart of Home Assistant during a charging session can give a fake spike in the logged
   consumption that needs to be manually edited using "Adjust sum" in the Statistics tab of the Developer tools dashboard.
 
+## Features missing from the API
+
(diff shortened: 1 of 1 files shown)


sveinse (approved):
Excellent. Approved. Thank you.

Issue #412, opened 2026-07-25
Installation address and charger name are logged in plaintext despite `REDACT_LOGS`


enoch85:
Disclosure, AI helped me write this.

## Environment

- Zaptec integration version: `0.8.7`
- Home Assistant: 2026.7.1 (Container, Docker)
- Charger: Zaptec Go (Apollo), firmware 2.6.0.6
- `REDACT_LOGS = True` (default)
- Log level: `custom_components.zaptec: debug`

## The problem

The redaction database correctly maps the installation address and the charger
name to placeholders, and `>>> Adding` / `>>> Ignoring` lines honour that:

```
>>>  Adding   Installation[437ede].address (Address)  =  <str> '<--Redact #2-->'
>>>  Adding   Charger[d3b1b9].name (Name)  =  <str> '<--Redact #6-->'
```

But the polling lines print the same values unredacted, in parentheses after
the qualified id:

```
Poll info from Installation[437ede] (<my street address>)
Poll firmware info from Installation[437ede] (<my street address>)
Poll state from Charger[d3b1b9] (<my charger serial>)
```

The `qual_id` part is redacted; the display name in parentheses is not.

These lines repeat on every poll cycle, so with debug logging enabled a log
contains hundreds of copies of the user's home address — even though the
`Redaction database` block at the end shows it as `<--Redact #2-->` and the
user has no reason to think it leaked.

## Secondary: charger UUID in exception messages

Exceptions are raised with the raw URL, bypassing redaction entirely:

```
custom_components.zaptec.zaptec.exceptions.RequestError: POST request to
https://api.zaptec.com/api/chargers/<full-charger-uuid>/SendCommand/10001
failed with status 500: <ClientResponse(...)>
```

The same UUID appears redacted as `<--Charger[xxxxxx]-->` everywhere else.

## Why this matters

The redaction feature exists so users can share debug logs on GitHub issues.
A user who checks the `Redaction database` block, sees everything replaced by
placeholders, and strips that block before posting will still publish:

- their home address (many times over)
- their charger serial number
- the charger's real UUID, if any request 500'd

This is easy to miss precisely because the redaction machinery looks like it
is working.

## Expected behaviour

Display names and URLs in log output pass through the same `redact()` helper
already used for `>>> Adding` lines, so that with `REDACT_LOGS = True` no
identifying value appears in plaintext anywhere in the log.

## Suggested fix

- Route the `(name)` suffix in the `Poll …` log calls through `redact()`.
[...]

PR #413, opened 2026-07-26
Increase redaction from polling calls and request-errors

steinmn:
Addresses #412

### DIFF
--- a/custom_components/zaptec/zaptec/api.py
+++ b/custom_components/zaptec/zaptec/api.py
@@ -298,7 +298,9 @@ class Installation(ZaptecBase):
 
     async def poll_info(self) -> None:
         """Update the installation info."""
-        _LOGGER.debug("Poll info from %s (%s)", self.qual_id, self.get("Name"))
+        _LOGGER.debug(
+            "Poll info from %s (%s)", self.qual_id, self.zaptec.redact(self.get("Name"))
+        )
 
         # Get the installation data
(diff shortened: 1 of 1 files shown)


custom_components/zaptec/zaptec/api.py
```diff
@@ -1211,7 +1219,7 @@ async def request(
                     return json_result
 
                 error = RequestError(
-                    f"{method.upper()} request to {full_url} failed with status {response.status}: {response}",
+                    f"{method.upper()} request to {self.redact(full_url)} failed with status {response.status}: {response}",
```
sveinse, line 1222: I'm conflicted about this. This makes the URL unusable for any further debugging. Specifically one must rerun the integration with redaction disabled to get the proper address. Do you think that's an acceptable compromise to protect the UUID?

steinmn: I mean, we already instruct the user to reload the integration in the second paragraph of the [debugging-section of the readme](https://github.com/custom-components/zaptec?tab=readme-ov-file#debugging), so then you can just find the appropriate value in the redaction list. I'm also a fan of consistency, and we redact the same in the sending of the request :shrug:

sveinse: Yes, you're right. Let's go with this

sveinse (approved):
lgtm

enoch85:
Thank you for the quick turnaround! :tada:

Issue #257, opened 2025-07-29
Cleanup directory structure and split between API library and HA integration


sveinse:
The access library (`api.py` and friends) should be split off from the HA integration parts (`__init__.py` and co). In HA integrations these two components are kept separate, but in Zaptec these are merged as one.

If ever zaptec is included into HA core, the API access parts will have to be a separate library on pypi. For that I'm thinking the namespace `zaptec`, such as `from zaptec import Zaptec`. To prepare for this, I think we should create `zaptec/` to add them into.

The HA parts needs some cleanup too. `__init__.py` is very large, and it is customary to have `entity.py` and `coordinator:py`. We should split the `__init__.py` into these files.

This issue should be executed with absolutely minimal code changes, as this will create a lot of diff changes. Any changes to the code will be hidden in these moves.

-[ ] Move `api.py` and associated files into `zaptec/`
-[ ] Split `__init__.py` into `entity.py` and `coordinator.py`

PR #414, opened 2026-07-26
test: migrate coordinator/entity tests to pytest-homeassistant-custom-component harness

rhammen:
## Summary

Replaces #394. Maintainer review on #394 (CHANGES_REQUESTED) asked how gold/platinum HA integrations test coordinators and entities — the answer is the real `pytest-homeassistant-custom-component` harness (real `hass`, `MockConfigEntry`, behavior asserted via `hass.states`), not hand-rolled `FakeConfigEntry`/mock-`hass` doubles poking private attributes.

- Adopts pytest-hacc for `tests/test_*.py` (coordinator/entity/init), pinned per-Python via environment markers to match each CI leg's HA version.
- **Scopes the harness to those tests only** (`pytest tests --ignore=tests/zaptec`), leaving `tests/zaptec/*` — the vendored `zaptec/` API client, a standalone-PyPI-library-in-waiting per #257 — running as plain pytest (`-p no:homeassistant`), so its live `api.zaptec.com/api/constants` call keeps working and the two concerns stay separated as #257 intends. Two pytest invocations, combined via `--cov-append`.
- Deliberately **Linux-native**: no native-Windows accommodation (no committed shim, no `pytest_plugins` force-load). On Linux, pytest-hacc autoloads on its own — CI and the devcontainer both just work.
- `coordinator.py` and `entity.py` coverage: **100% / 100%** (up from ~27-29% pre-migration).

## Sequencing note

#402, #401, #399, #397 each carry their own copy of `tests/conftest.py`/`test_coordinator.py`/`test_entity.py`/`test_init.py` using the old `FakeConfigEntry` pattern, built independently of this branch. I've marked all four as draft for now. Once this merges, each will either be updated in place to the new test conventions and re-marked ready, or - if that turns out easier - replaced with a new PR off a fresh branch, the same way #396 was replaced by the `feature/energy-statistics-standalone` branch.

## Test plan

- [x] `pytest tests --ignore=tests/zaptec --cov=./custom_components/zaptec --cov-branch` - 24 passed, 1 skipped
- [x] `pytest tests/zaptec -p no:homeassistant --cov=./custom_components/zaptec --cov-branch --cov-append` - 102 passed, 1 skipped (`--skip-api`)
- [x] `coordinator.py` 100%, `entity.py` 100%
- [x] Verified in both the maintainer-recommended Dev Container and CI (fork, both Python matrix legs)
- [x] `ruff format --diff` / `ruff check` clean
- [x] No `custom_components/**` production changes - test/infra-only

## AI policy compliance

I personally reviewed this code together with Claude to the best of my knowledge. That review resulted in several updated docstrings for things that weren't clear or obvious to me on first read. Based on that process, I believe this complies with the AI policy to the best of my ability.


scripts/test
```diff
@@ -5,8 +5,14 @@ set -e
 if [ "$1" == "--skip-api" ]; then
     export SKIP_ZAPTEC_API_TEST="true"
 fi
+
+# HA-integration tests run under the pytest-hacc harness (autoloads on Linux).
+# API-client tests (tests/zaptec/*, the future-standalone client per #257) run
+# as plain pytest with the harness disabled, so their live constants call is not
+# socket-blocked. Coverage from both is combined via --cov-append.
```
sveinse: Be mindful about usage of issue references in the code. The code will likely move to other repos and then the issues will have no reference (which coincidentally is what is being talked about here).

tests/conftest.py
sveinse, line 1: Since the split between the API parts and the HA integration is imminent, please be advised to try to keep a clean separation between what setup belongs to which of these two. I propose to make a `tests/zaptec/contest.py` file and place the API specific test setup there.

requirements.txt
```diff
@@ -6,5 +6,5 @@ ruff==0.15.22
 # Copy from manifest.json to get this into the dev container
 # without needing to start HA
 azure-servicebus==7.14.3
-pydantic==2.13.1
+pydantic>=2.11.7,<2.14
```
sveinse, line 9: Why the downgrade of pydantic?

requirements_test.txt
```diff
@@ -1,4 +1,14 @@
...
+# pytest-homeassistant-custom-component pins an EXACT `homeassistant==`, so the
+# release must match the HA that requirements.txt installs on each CI Python leg
+# (validate.yaml sed-reverts HA to 2026.2.3 on 3.13). Select the matching release
+# per Python version: newest releases require Python >=3.14, so 3.13 uses the last
+# 3.13-compatible one. 0.13.324 -> HA 2026.4.3 (py>=3.14); 0.13.316 -> HA 2026.2.3.
+# It also brings the full version-matched pytest stack, and pins pydantic to HA's
+# version (2.12.2), which is why requirements.txt uses a pydantic RANGE (matching
+# manifest.json) rather than an exact pin that would conflict.
+pytest-homeassistant-custom-component==0.13.324; python_version >= "3.14"
+pytest-homeassistant-custom-component==0.13.316; python_version < "3.14"
```
sveinse: @steinmn Is it time for us to ditch support for python 3.13 in favor of 3.14? I believe the HA version we target is using no less than 3.14.

sveinse (commented):
The introduction of two runs of pytest was unexpected. What is the main reason for not running them all in one? I.e. why can't we run the API tests under the HA test bench? -- Since the split is imminent, its not necessarily a bad thing to do it like this. But it does create more code and more steps, so we need a clear reason why not run it in one.

Another observation is that the AI assistance creates _a lot_ of comments in the code. While it is good to be informed what the full intent is in the different functionds, thay are very wordy. A lot more wordy than any developer would do. And at some point the blocks of text disrupts the flow of the code. Is it possible to instruct the agent to reduce the words?

rhammen:
> The introduction of two runs of pytest was unexpected. What is the main reason for not running them all in one? I.e. why can't we run the API tests under the HA test bench? -- Since the split is imminent, its not necessarily a bad thing to do it like this. But it does create more code and more steps, so we need a clear reason why not run it in one.

pytest-hacc's hass fixture blocks all non-localhost sockets. tests/zaptec/test_zconst.py/test_redact.py need a real call to api.zaptec.com/api/constants (deliberate — catches drift against the live API), so they can't run under the harness at all.

I did look at running everything in one invocation via a snapshot fixture instead of the live call, but that just trades a real API-drift check for a fake one to avoid the split, so I dropped it.

Given #257 already commits zaptec/ to becoming a standalone package tested independently of HA, and splitting themm allows one test to run under the harness and the other one to do the live api call, I judged that the better, cleaner solution.

> Another observation is that the AI assistance creates _a lot_ of comments in the code. While it is good to be informed what the full intent is in the different functionds, thay are very wordy. A lot more wordy than any developer would do. And at some point the blocks of text disrupts the flow of the code. Is it possible to instruct the agent to reduce the words?

Yes of course. I'm still a bit searching for the optimum here. 
The AI policy urges me to review and understand the code. In such review, I encountered choices the AI made that were not obvious to me, and the AI also judged some of them as non obvious. That is where I made the call to add the explanation to the docstrings.... So that was my call. I can skip that step next time, or ask the AI to be more concise, but then be kind on me regarding the AI policy. 😄

requirements.txt
rhammen: Not a downgrade of the target, just dropping the exact pin. `manifest.json` already declares `pydantic>=2.11.7,<2.14` (unchanged here) — `requirements.txt` previously hard-pinned `==2.13.1`, stricter than that range.

pytest-hacc pins an exact HA release, and HA's dependency chain resolves to an exact pydantic of its own — `2.12.2` for the release this PR's Python-3.14 leg installs. `scripts/setup` installs `requirements.txt` and `requirements_test.txt` together, so they have to co-resolve: `==2.13.1` alongside a transitive `==2.12.2` is simply uninstallable. Relaxing to the manifest range fixes that, and confirmed locally — resolved pydantic is `2.12.2`, inside the range.

scripts/test
rhammen: Agreed. I have remove this reference and 3 others from the code.

sveinse:
> > Another observation is that the AI assistance creates _a lot_ of comments in the code. While it is good to be informed what the full intent is in the different functionds, thay are very wordy. A lot more wordy than any developer would do. And at some point the blocks of text disrupts the flow of the code. Is it possible to instruct the agent to reduce the words?
> 
> Yes of course. I'm still a bit searching for the optimum here. The AI policy urges me to review and understand the code. In such review, I encountered choices the AI made that were not obvious to me, and the AI also judged some of them as non obvious. That is where I made the call to add the explanation to the docstrings.... So that was my call. I can skip that step next time, or ask the AI to be more concise, but then be kind on me regarding the AI policy. 😄

That is a very good approach for gaining the necessary understanding to be the "developer in the loop" before submitting the PR. As with the superpower docs you're using, I think they can help you to gain the required understanding of the problem and the fix. Perhaps a good way would be to generate the code in its full comment verbosity first so you can understand it, and then instruct the agent to reduce and compact the comments? I think that could work.

In case you encounter something you don't understand, ask the agent to clarify. This is one of its great strengths. 👍

requirements_test.txt
steinmn: I think dropping 3.13 support for 0.9.0 makes sense. Since this also means we would be dropping support for ha<2026.3, I think waiting for the minor version update would be preferred.

I think the code speaks for itself here, so the comment block can be removed.

tests/conftest.py
rhammen: Agreed, done — added `tests/zaptec/conftest.py` with `zaptec_constants` (the only fixture used exclusively by `tests/zaptec/*`).

`zaptec_username`/`zaptec_password` stay in the root conftest for now: they're also needed by `tests/test_diagnostics.py`, which isn't under `tests/zaptec/` and isn't converted to the new pattern yet (that's #402, currently draft and dependent on this PR). Pytest fixtures only resolve looking up the directory tree, not sideways, so moving those two now would break `test_diagnostics.py`'s collection until #402 lands. Once that conversion's in, happy to finish the move.

requirements_test.txt
rhammen: Comment block removed, the two per-Python lines stay as-is.

rhammen:
Trimmed the code comments across the branch — cut every docstring/comment down to the non-obvious "why", dropped the restated "what" and secondary asides. Net -66 lines in the test files, same coverage/behavior.

PR #415, opened 2026-07-27
Relax over-strict validation models to fields api.py uses (fixes #359, #400)

### LINKED ISSUE #359: Review all validation models and make sure we don't require something we don't use.

### LINKED ISSUE #400: Setup fails because ChargerFirmware requires fields no longer returned by the API
My setup is:
- Home Assistant Core: 2026.7.1
- Home Assistant OS: 18.1
- Supervisor: 2026.06.2
- Frontend: 20260624.4
- Zaptec integration: 0.8.7

The failure is dues to ChargerFirmware does not contain the fields `CurrentVersion `and `IsUpToDate`. Making those optional locally solves the issue for me.

rhammen:
Supersedes #397 (un-stacked: this branch is off `master`, no test-PR ancestry). Independent of #414.

Several validation models required response fields `api.py` never hard-reads, so a reduced or slimmed-down API response failed setup instead of degrading. Rule applied: a field is required only if `api.py` reads it via a hard subscript (`d["Key"]`); everything else is optional.

- **`ChargerFirmware`** — only `ChargerId` required; `CurrentVersion`/`IsUpToDate`/etc. now optional. Fixes #400.
- **`Circuit.MaxCurrent`** — was hard-indexed in `Installation.build()` but never declared, so a response missing it crashed with a raw `KeyError`. Now validated; `build()` also tolerates null `Chargers`/`Name`.
- **`Installation`/`Charger`/`Hierarchy`** — role-dependent fields relaxed to optional; only the hard-read keys (`Id`, `DeviceType`) stay required.

Incorporates review feedback from #397: reuses `Charger` for a Circuit's chargers instead of a separate model, keeps `Hierarchy.Id` required.

Tested against a real installation across four account roles (no `ValidationError`), plus offline unit tests.

I reviewed this code myself and believe it complies with the AI policy.

Fixes #359
Fixes #400

### DIFF
--- a/custom_components/zaptec/zaptec/api.py
+++ b/custom_components/zaptec/zaptec/api.py
@@ -274,14 +274,15 @@ class Installation(ZaptecBase):
             redact.add_uid(ctid, "Circuit")
             _LOGGER.debug("    Circuit %s", redact(ctid))
 
-            for charger_item in circuit["Chargers"]:
+            # Chargers and Name are nullable per the Zaptec API docs.
+            for charger_item in circuit.get("Chargers") or []:
                 chgid = charger_item["Id"]
                 redact.add_uid(chgid, "Charger")
 
                 # Inject additional attributes
                 charger_item["InstallationId"] = self.id
                 charger_item["CircuitId"] = ctid
-                charger_item["CircuitName"] = circuit["Name"]
+                charger_item["CircuitName"] = circuit.get("Name")
                 charger_item["CircuitMaxCurrent"] = circuit["MaxCurrent"]
 
                 # Add or update the charger
--- a/custom_components/zaptec/zaptec/api.py
+++ b/custom_components/zaptec/zaptec/api.py
@@ -269,13 +269,24 @@ class Installation(ZaptecBase):
         redact = self.zaptec.redact
 
         self.chargers = []
+        if not hierarchy.get("Circuits"):
+            _LOGGER.warning("Installation %s contains no circuits.", self.qual_id)
+            return
+
         for circuit in hierarchy["Circuits"]:
             ctid = circuit["Id"]
             redact.add_uid(ctid, "Circuit")
             _LOGGER.debug("    Circuit %s", redact(ctid))
 
-            # Chargers and Name are nullable per the Zaptec API docs.
-            for charger_item in circuit.get("Chargers") or []:
+            if not circuit.get("Chargers"):
+                _LOGGER.warning(
+                    "Circuit %s of installation %s contains no chargers.",
+                    redact(ctid),
+                    self.qual_id,
+                )
+                continue
+
+            for charger_item in circuit["Chargers"]:
                 chgid = charger_item["Id"]
                 redact.add_uid(chgid, "Charger")
 
--- a/custom_components/zaptec/zaptec/validate.py
+++ b/custom_components/zaptec/zaptec/validate.py
@@ -16,10 +16,10 @@ class Installation(BaseModel):
 
     model_config = ConfigDict(extra="allow")
     Id: str
-    Active: bool
-    CurrentUserRoles: int
-    InstallationType: int
-    NetworkType: int
+    Active: bool | None = None
+    CurrentUserRoles: int | None = None
+    InstallationType: int | None = None
+    NetworkType: int | None = None
 
 
 class Installations(BaseModel):
@@ -27,7 +27,7 @@ class Installations(BaseModel):
 
     model_config = ConfigDict(extra="allow")
     Data: list[Installation]
-    Pages: int
+    Pages: int | None = None
 
 
 class Charger(BaseModel):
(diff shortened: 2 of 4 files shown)


custom_components/zaptec/zaptec/api.py
```diff
@@ -274,14 +274,15 @@ async def build(self) -> None:
             redact.add_uid(ctid, "Circuit")
             _LOGGER.debug("    Circuit %s", redact(ctid))
 
-            for charger_item in circuit["Chargers"]:
+            # Chargers and Name are nullable per the Zaptec API docs.
+            for charger_item in circuit.get("Chargers") or []:
```
steinmn: Having Chargers actually be null/None seems absurd, and potentially a sign of an installation that is incorrectly set up. With this setup we pass it by silently. I think we should log a warning instead.

Additionally, Circuits is also nullable (even more absurd), so if we're going down this road, we should guard against that as well.

So something like
```
        if not hierarchy.get("Circuits"):
            _LOGGER.warning(
                ("Installation %s contains no circuits."),
                self.qual_id,
            )
            return

        for circuit in hierarchy["Circuits"]:
            ctid = circuit["Id"]
            redact.add_uid(ctid, "Circuit")
            _LOGGER.debug("    Circuit %s", redact(ctid))

            if not circuit.get("Chargers"):
                _LOGGER.warning(
                    ("Circuit %s of installation %s contains no chargers."),
                    redact(ctid),
                    self.qual_id,
                )
                continue

            for charger_item in circuit["Chargers"]:
```

custom_components/zaptec/zaptec/validate.py
```diff
@@ -16,27 +16,27 @@ class Installation(BaseModel):
...
 
     model_config = ConfigDict(extra="allow")
     Data: list[Installation]
-    Pages: int
+    Pages: int | None = None
 
 
 class Charger(BaseModel):
-    """Pydantic model for a Zaptec charger."""
+    """Pydantic model for a Zaptec charger, as returned by /chargers and /chargers/{id}."""
```
steinmn: no need for the as-returned-part

custom_components/zaptec/zaptec/validate.py
```diff
@@ -61,17 +61,21 @@ class Circuit(BaseModel):
 
     model_config = ConfigDict(extra="allow")
     Id: str
-    Name: str
-    Chargers: list[Charger]
+    # Nullable per the Zaptec API docs; api.py reads it defensively via .get().
+    Name: str | None = None
+    MaxCurrent: float
+    # Reuses Charger: a hierarchy-only charger is never re-merged with /chargers,
+    # so its required DeviceType (hard-subscripted later in build()) comes only from here.
```
steinmn: Leftover comment from when there was a separate hierarchy Charger model?

custom_components/zaptec/zaptec/validate.py
```diff
@@ -61,17 +61,21 @@ class Circuit(BaseModel):
 
     model_config = ConfigDict(extra="allow")
     Id: str
-    Name: str
-    Chargers: list[Charger]
+    # Nullable per the Zaptec API docs; api.py reads it defensively via .get().
```
steinmn: Don't think this comment is needed

steinmn (changes requested):

custom_components/zaptec/zaptec/api.py
rhammen: Agreed — added the warn-and-skip guard for both Circuits and Chargers as suggested, in 2359337.

custom_components/zaptec/zaptec/validate.py
rhammen: Agreed, reverted to the plain docstring in 2359337.

custom_components/zaptec/zaptec/validate.py
rhammen: Correct, it was leftover — removed in 2359337.

custom_components/zaptec/zaptec/validate.py
rhammen: Agreed, removed in 2359337.

sveinse (approved):
I tested this on my dev setup using my test user that have no access and it works fine. For what I can see, this looks good. Approved.

sveinse:
@steinmn do you concur that this is ready? You requested changes and must reapprove

steinmn (approved):
LGTM

PR #416, opened 2026-07-27
Add per-charger energy statistics feed for Energy Dashboard

rhammen:
Implements feature #300.

The live *Energy Meter* sensor is polling-delayed, so energy lands in the wrong hour on the Energy Dashboard (see #300, also #162). This adds a separate long-term-statistics feed per charger, backfilled from Zaptec's archived charge sessions and imported hourly, so historical consumption is attributed to the correct hour.

- New `ZaptecStatisticsCoordinator` (statistics.py) imports `/api/sessions/archived` into HA external statistics as `zaptec:energy_<id>`, shown in the Energy Dashboard device picker as "<charger> Energy".
- Recorder is an `after_dependency`: if the recorder is disabled the integration still loads, only this feed is skipped.
- Requires the Owner role (the endpoint is Owner-only); a 403 degrades just this feed with a warning, not the whole integration.
- Backfills up to 730 days on first run, then imports hourly.

Because it is built from completed sessions, the most recent hour or two can lag — it's for accurate historical graphs, not real-time (the live sensor still covers that).

Tests cover the bucketing logic and the archived-sessions request. Full harness-based coordinator tests will follow once the test-framework migration (#414) lands.

This was written with AI assistance; I have personally reviewed the code. I've also tried to trim the docstrings/comments down in line with maintainer preference — happy to cut further if any still read as too verbose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,18 @@ Confirmed to work with Zaptec products
   [here](https://www.home-assistant.io/common-tasks/general/#defining-a-custom-polling-interval),
   will have unexpected effects. If the automatic polling is turned off, not all
   the data in the integration will update properly.
-* Using the _Energy Meter_ entity as an input to the Energy Dashboard will give values that are delayed by 1 hour
-  in the graphs (see [issue 162](https://github.com/custom-components/zaptec/issues/162) for details).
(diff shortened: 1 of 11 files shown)


sveinse:
As this PR is new functionality, I'm inclined to postpone this to 0.9.0.

PR #418, opened 2026-07-28
Reconnect the live update stream after a transient failure

### LINKED ISSUE #417: Live update stream doesn't reconnect after a transient connection failure (e.g. AMQP/network drop)
## Description

When the live Service Bus (AMQP) stream connection fails — e.g. due to a
transient home-internet outage — the integration doesn't reconnect. It
requires a full integration/HA reload to restore live push updates.

## Observed error

```
Deze fout is ontstaan door een aangepaste integratie.

Logger: custom_components.zaptec.zaptec.api
Bron: custom_components/zaptec/zaptec/api.py:433
Integratie: Zaptec EV charger (documentatie, problemen)
Eerst voorgekomen: 26 juli 2026 om 05:01:31 (1 gebeurtenis)
Laatst gelogd: 26 juli 2026 om 05:01:31

Stream failed
  File "/config/custom_components/zaptec/zaptec/api.py", line 433, in stream_main
    async for msg in receiver:
    ...<33 lines>...
        await receiver.complete_message(msg)
  File "/usr/local/lib/python3.14/site-packages/azure/servicebus/aio/_transport/_pyamqp_transport_async.py", line 234, in iter_contextual_wrapper_async
    message = await receiver._inner_anext(wait_time=max_wait_time)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/azure/servicebus/aio/_servicebus_receiver_async.py", line 236, in _inner_anext
    return await self._do_retryable_operation(self._iter_next, wait_time=wait_time)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/azure/servicebus/aio/_base_handler_async.py", line 261, in _do_retryable_operation
    raise last_exception from None
azure.servicebus.exceptions.ServiceBusConnectionError: Failed to initiate the connection due to exception: [Errno -3] Try again Error condition: amqp:socket-error.
```

I suspect this was a temporary internet disconnect — my modem/router
likely rebooted around that time.

## Root cause

`stream_main()` catches the exception, logs "Stream failed", and exits
via its `finally` block (setting `_stream_running = False`) without
reconnecting:

https://github.com/custom-components/zaptec/blob/master/custom_components/zaptec/zaptec/api.py#L474-L482

`create_streams()` in `manager.py` starts the stream task exactly once,
during `async_setup_entry`, with nothing supervising or restarting it if
it dies:

https://github.com/custom-components/zaptec/blob/master/custom_components/zaptec/manager.py#L193-L205

## Impact

Not a total outage — the polling coordinator keeps running independently
on its own interval, so state isn't completely stale. But the live/instant
push updates are silently lost until the user notices and reloads the
integration or restarts Home Assistant. There's no warning that the stream
is down beyond the one-time "Stream failed" log entry.

## Suggested fix

Wrap `stream_main()` in a reconnect loop (with backoff) so a transient
connection failure is retried automatically, similar in spirit to the
REST-request retry logic added in #393.

rhammen:
Fixes #417.

The live update stream (Service Bus/AMQP connection) currently dies
permanently after any transient failure (e.g. a brief internet outage) —
it swallows the exception, logs it, and just stops, requiring a full
integration reload to recover.

- `stream_main()` no longer catches its own exceptions; only the
  existing 403/Forbidden case still returns cleanly ("permanent stop").
- New `_stream_supervisor()` in `manager.py` wraps `stream_main()` in a
  reconnect loop with exponential backoff + jitter, retrying any
  transient failure indefinitely while leaving `CancelledError` (task
  cancellation on unload/reload) unaffected.
- New `STREAM_RECONNECT_*` constants in `const.py`.

Personally reviewed the code changes and trimmed AI-generated
docstrings/inline comments for length.


.gitignore
```diff
@@ -11,6 +11,7 @@ __pycache__
 coverage.xml
 .ruff_cache
 htmlcov/
+.superpowers/
```
sveinse: Please remove this. This is not something most ppl use. You can add an exception in your local git config (use claude to help you set that up).

custom_components/zaptec/manager.py
```diff
@@ -32,6 +43,45 @@ class ZaptecEntityDescription(EntityDescription):
...
+                )
+                warned = True
+            else:
+                _LOGGER.debug("Stream for %s still reconnecting", install.qual_id, exc_info=True)
+            await asyncio.sleep(delay)
+            delay = min(delay * STREAM_RECONNECT_FACTOR, STREAM_RECONNECT_MAX_DELAY)
+            delay = random.normalvariate(delay, delay * STREAM_RECONNECT_JITTER)
+            delay = min(delay, STREAM_RECONNECT_MAX_DELAY)
+        else:
+            return
```
sveinse, line 103: Item 1) I would love to have a message like "Connected to stream after %s reconnects". That mirrors well with the first message "Stream for % disconnected, reconnecting". But thinking about how to solve that might not be that easy.

Item 2) The `warned` flag is `False` by default and will be set to `True` if an exception occurs. If the stream succeeds and then runs for a long time until failing, this will still be considered a reconnect. Once it has reconnected it will stay `True` forever. Perhaps there should be something that resets the `warned` flag?

custom_components/zaptec/zaptec/api.py
```diff
@@ -471,10 +471,6 @@ async def stream_main(
                         # remove the msg from the "queue"
                         await receiver.complete_message(msg)
 
-        except Exception:
-            # Do this in order to show the error in the log.
-            _LOGGER.exception("Stream failed")
-
```
sveinse, line 477: By taking away this exception logging, there is no traceback stored in the debug output anywhere. The new `_stream_supervisor()` does not print the exception.

Those users that experience issues with the stream handling, what kind of exceptions do they get? Perhaps we can except those exceptions with a general "stream disconnected" message, but print the full traceback for anything unexpected?

sveinse (changes requested):

sveinse:
@rhammen Is this a fix to #423?

.gitignore
rhammen: Removed — it's in my local `.git/info/exclude` now.

custom_components/zaptec/manager.py
rhammen: Both items done.

Item 1: `stream_main()` takes an optional `on_connect` callback, fired once the receiver is open. The supervisor uses it to log `Stream for %s connected after %s reconnect attempt(s)`, only after an actual reconnect — a normal first start stays quiet.

Item 2: `warned` is replaced by a `reconnects` counter, reset from that same callback. The old condition measured time since the attempt *started*, so an attempt that hung for 300 s without ever connecting would have reset it. Now it takes a real connection that then stays up for `STREAM_RECONNECT_STABLE_TIME` (300 s).

I measured my own installation to pick that value: while charging, observations arrive with a median gap of 164 s (35 updates over 100 min); idle, it's a two-message burst every ~68 min. I first used 60 s, which is below the charging cadence — a connection could be called healthy having carried nothing, and a stream flapping just above 60 s would reset the backoff and warn again every cycle. It ends up equal to `STREAM_RECONNECT_MAX_DELAY`, but the two answer different questions so I kept them separate.

custom_components/zaptec/zaptec/api.py
rhammen: Done, with the split you suggested. `STREAM_TRANSIENT_ERRORS` — `ServiceBusError`, `MessageAlreadySettled`, `OSError`, `TimeoutError`, `aiohttp.ClientError`, `ZaptecApiError` — warns terse and names the exception; anything else keeps the full traceback at warning level. Repeat failures within one outage log the traceback at debug.

Two of those need a word. `MessageAlreadySettled` derives from `ValueError`, not `ServiceBusError`, and `complete_message()` sits outside the per-message try/except, so a lock expiry escapes. `ZaptecApiError` is there because a non-403 `RequestError` from `live_stream_connection_details()` propagates too — that's the #417 scenario itself, so it belongs in the expected bucket.

As for which exceptions users actually get: #417 has no traceback attached, so this list is what the stack can plausibly raise rather than a confirmed sample. Anything outside it now shows up loudly, which should tell us.

One correction to the premise: `_stream_supervisor()` did already log with `exc_info=True`, so the traceback wasn't lost — but the signal/noise split you're describing is better, hence the change.

rhammen:
No — #423 looks like an account-side problem rather than a code bug, so this PR won't fix it.

What it does change for that failure: `RequestRetryError` now propagates as a transient error, so the stream retries with backoff instead of dying on the first failure. If a `messagingConnectionDetails` failure is temporary, that self-heals. #423's isn't — I checked the portal today and the reporter's messaging subscription appears not to be provisioned (their "subscription enabled" toggle is greyed out where mine is on), which no amount of retrying will fix. Details in #423.

One consequence worth your view before this merges. For a permanently failing stream, the supervisor now warns once and then logs only at debug, retrying every 5 minutes forever. That's right for a transient outage but hides a terminal one — a #423-style user would see a single warning and then silence. Should it re-warn periodically (say hourly) while an outage persists?

rhammen:
Hit #417 in the wild yesterday, so this now has a field reproduction rather than
just a theoretical one.

The stream had stopped delivering some time before 15:10 and was still silent at
23:27 — over eight hours, with no reconnect attempt. Everything kept working on
the poll interval, so nothing looked broken. Confirmed with debug logging: zero
stream frames across a window containing a cable plug-in and an RFID tap, while a
forced `homeassistant.update_entity` picked both up at once. A
`reload_config_entry` logged `Running service bus stream for Installation[...]`
and frames resumed within seconds:

    23:30:32.960  710 (ChargerOperationMode) = 1   (cable unplugged)
    23:31:04.743  710 (ChargerOperationMode) = 2   (plugged back in)

Not a #423-style failure: `messagingConnectionDetails` succeeded, no 403, and all
entities were present throughout.

It doesn't add an exception sample though — debug logging only went on hours after
the stream had already stopped, so the failure itself isn't in any log I have.
`STREAM_TRANSIENT_ERRORS` is still an educated list rather than a measured one.

It does narrow my re-warn question above: this outage would have self-healed. The
reload reconnected immediately, so the supervisor would have succeeded on the first
or second attempt, logging one warning and then the "connected after N attempt(s)"
line. Periodic re-warning only matters for a stream whose retries never succeed —
a different failure from this one, and one where the reporter is already looking at
their logs. So I'd leave it as is, and drop the question unless you'd rather have it.

Issue #200, opened 2025-07-17
Feature request : Identify which rfid tag is used to start a session


Lheriss:
Hi,
As asked, I create a new issue / feature request
I would like to have a way to identify which rfid tag is used to start a session. I have configured three of them, and would like to be able to detect how much energy is used with each of them...
Many thanks !

sveinse:
Thanks for looking at it. If there is no immediate burning desires for getting it included, then I agree. Let's wait after v0.8 is out. We have enough

sveinse:
@Lheriss While waiting for the implementation of this feature, it could be valuable to know that all data-points from Zaptec is available a s attributes in the "Conneted" object of the charger and the integration. It can be accessed via a blueprint: https://github.com/custom-components/zaptec?tab=readme-ov-file#templates

sveinse:
The NFC is available in attribute `charger_currrent_user_uuid`. If the charging was authenticated by NFC it will read `nfc-xxxxxxxxxxxxxx` where `x` is the NFC ID which has been added to the user account. If using the "Authorize charging" it will be `ble-xxxxxxxx-bbbb-cccc-dddd-000000000000`

sveinse:
In 0.9.0 we plan to remove the attributes, so I think the authentication information, e.g. which NFC tag, should be made a full entity.

rhammen:
#425 covers the first half of this: `charger_current_user_uuid` never cleared, so it kept the previous session's token — which is why it disagreed with `AuthenticationCode` in @nelgmo's report above.

The sensors themselves are ready on top of that: "Authorized by" (observation 722) and "Last session authorized by" (`completed_session.AuthenticationCode`), with `ble-` tokens shown as a remote authorization per @sveinse's note about the `nfc-`/`ble-` prefixes.

Implemented locally. Will test on my next charging session to confirm behavior.

MelchMwoan:
Hey @rhammen, how is this coming along? 
I see there is a PR open for it but that hasn't been merged yet, is it broken or untested?
You need any help with testing?

I'd love to be able to get the RFID tag of the user that authorized the charging session.

rhammen:
Hi @MelchMwoan ,
From my perspective, it is ready and waiting to be merged. 
Merging needs to be done (approved) by a maintainer, I cannot do that, so we have to wait for that. 
If you want to test, I have a branch in my fork that contains ALL open PRs on this repo. Feel free to try and test that one:
[...]

PR #425, opened 2026-08-26
Clear session observations that Zaptec sends without a value

rhammen:
Zaptec ends a charging session by sending observations 721 (`SessionIdentifier`) and 722 (`ChargerCurrentUserUuid`) with no `Value`/`ValueAsString` field at all. `state_to_attrs` skips entries without a value, so the attributes kept the previous session's values indefinitely.

This explains the report in [#200 (comment)](https://github.com/custom-components/zaptec/issues/200#issuecomment-3989612118): `charger_current_user_uuid` appearing not to update with the RFID tag, while `AuthenticationCode` did. The attribute wasn't failing to update — it was never cleared, so it kept showing the previous session's token.

Map those two ids to an empty value instead of skipping them. Scoped to 721/722: `Pulse` is the only other observation seen without a value, and it never carries one.

Verified on a Zaptec Go 2 (fw 3.3.0.0): both observations appear valueless in the charger state after a session ends, and the derived attributes now clear.

Part of #200; the "Authorized by" sensors follow in a separate PR on top of this.

### DIFF
--- a/custom_components/zaptec/zaptec/api.py
+++ b/custom_components/zaptec/zaptec/api.py
@@ -32,6 +32,7 @@ from .const import (
     API_TIMEOUT,
     API_URL,
     CHARGER_EXCLUDES,
+    CLEARABLE_OBSERVATIONS,
     DEFAULT_MAX_CURRENT,
     MAX_DEBUG_TEXT_LEN_ON_500,
     MISSING,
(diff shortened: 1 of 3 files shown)


rhammen:
Corrected the last line of the description: "within seconds of session end" was inferred from recorder history, which records when Home Assistant polled rather than when Zaptec sent the observation, so the timing was not something I actually measured. The valueless observations and the clearing behaviour are unchanged — only that claim was unsupported.

